PR #19, opened 2019-04-02
Update ansible-python-jobs to use fedora-latest-1vcpu

pabelanger:
We want python3 to be the default for testing, as a results we should be
using fedora-latest-1vcpu here. Another option is to default to
ubuntu-bionic, but that needs to be discussed.

### DIFF
--- a/zuul.d/project-templates.yaml
+++ b/zuul.d/project-templates.yaml
@@ -5,7 +5,9 @@
       A common set of python jobs to run.
     check:
       jobs:
-        - tox-linters
+        - tox-linters:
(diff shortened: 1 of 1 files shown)


pabelanger:
recheck

pabelanger:
recheck

pabelanger:
recheck

pabelanger:
recheck

pabelanger:
recheck

pabelanger:
recheck

pabelanger:
recheck

PR #30, opened 2019-04-11
Create ansible-network-vyos-base job

pabelanger:
This will be our base job for vyos related jobs. We are creating a 2
node nodeset with fedora-29 and vyos. We also modify our base job so it
will not run against the appliance group.

Depends-On: https://review.openstack.org/650880


pabelanger:
recheck

PR #77, opened 2019-05-21
Reset files for periodic vyos jobs

pabelanger:
We actually want to trigger jobs against all files.


pabelanger:
recheck

PR #85, opened 2019-05-22
Enable ARA for ansible-test integration testing

pabelanger:
It is helpful to give more info about ansible, in this case install ARA
and we will generate reports.


pabelanger:
recheck

PR #109, opened 2019-05-30
Increase testing of junos network-integration

pabelanger:
Now that we have PRs being created to fix issues, we should run all
junos tests.  stable branches are not reporting back to github, until
green.


pabelanger:
recheck

PR #110, opened 2019-05-30
Stop testing on stable-2.6 for network-integration

pabelanger:
This branch is in security mode, and even if we fixed issue, they likely
won't be merged. So rather then waisting test resources, drop testing.


pabelanger:
recheck

PR #118, opened 2019-06-06
Cap ARA<1.0.0 for ansible-test-network-integration

pabelanger:
We haven't looked to see what is needed to use ARA 1.0.0, so just cap it
for now.


pabelanger:
recheck

PR #137, opened 2019-06-15
Update iosxr file matcher to include netconf

pabelanger:
This is because we test netconf also with iosxr.


pabelanger:
recheck

PR #171, opened 2019-08-12
Add missing files in vyos jobs

danielmellado:
This commit adds some missing files for the vyos jobs, which weren't
triggered when enqueing a change.


danielmellado:
Pull-request updated, HEAD is now c21d3baebfc99cd39c14a7bf7ed546f68096e771

PR #192, opened 2019-08-26
Increase vyos platform timeout

danielmellado:
In order to test timeouts that had happened for vyos platform, this
commit increases vyos timeouts to 9000. Further issues would be fixed by
parallelizing rather than getting more increases on the timeout itself.


danielmellado:
That's just a test commit, so I added it to depends-on to check some stuff.
Feel free to disregard it. I'll add a [dnm,] tag on the title next time ;)

pabelanger:
Closing this out, as doesn't look to be needed

PR #195, opened 2019-08-29
Set ansible-collections-vyos-vyos template jobs voting

pabelanger:
Our vyos collection is passing properly now on integration testing. Time
to start voting in gate.


pabelanger:
recheck

PR #214, opened 2019-09-20
Switch network-integration jobs to use fedora-30

pabelanger:
Now that we have fedora-30 nodes online, we can migrate to them from
fedora-29.


pabelanger:
recheck

PR #253, opened 2019-11-28
add ansible-test-cloud-integration-vcsa-python36 job

goneri:
add ansible-test-cloud-integration-vcsa-python36 job
  
Add the new `ansible-test-cloud-integration-vcsa-python36` as a
non-voting job.

It only checks the integration roles that don't need any ESXi, just a
vcenter instance. It uses the new `zuul/vmware/vcenter_only` target
to target them.

There is still a couple of things that can be improved later:

- the vcenter appliance uses a vcenter.test FQDN, it should instead use
  origin VM hostname instead
- we reboot the vcenter to reload the configuration, this takes about 30s,
  and can probably be skipped
- in addition, there is several potential improvements that have been
  identified during the reviews.

### DIFF
--- a/playbooks/ansible-cloud/vcsa-appliance/files/bootstrap.yaml
+++ b/playbooks/ansible-cloud/vcsa-appliance/files/bootstrap.yaml
@@ -15,9 +15,19 @@
         state: present
         user: zuul
 
+    - name: Setup zuul user
+      authorized_key:
+        key: "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQCyiXfE1zHKdj6odbysr917Cn88ov0VQaPJtLKJyMNuRYAeMOFQHd50X8JO4dfZbmSo3YdJlVfz9FLRxE64mqj9bkN8hPFbkTG2F1AWXGPON5cmm4uiLPfQkWhX/LnClrhzZpNtMJYs5AEFeDs0POijcRugZsQA+wvLi0lSlhOfkqtjAJKpPUwy1wrJFDdvqdQBjpNQh/LB8c15XfQV2JT/3NX26dQe8zvHhL6NvfhBnAikodYkBr7UjSl36CBk0cPebZMZEBBiHdo76xORVkpmqDvkhFByXXeAsvRa2YWS4wxpiNJFswlRhjubGau7LrT113WMcPvgYXHYHf2IYJWD goneri@redhat.com"
+        state: present
+        user: zuul
+
+    - name: reset ssh connection
+      meta: reset_connection
+
 - hosts: appliance
   gather_facts: false
   tasks:
+
     - name: Generate random password
       set_fact:
         __password: "{{ lookup('password', '/dev/null length=15 chars=ascii_letters') }}"
--- a/playbooks/ansible-cloud/vcsa-appliance/files/bootstrap.yaml
+++ b/playbooks/ansible-cloud/vcsa-appliance/files/bootstrap.yaml
@@ -15,12 +15,6 @@
         state: present
         user: zuul
 
-    - name: Setup zuul user
-      authorized_key:
-        key: "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQCyiXfE1zHKdj6odbysr917Cn88ov0VQaPJtLKJyMNuRYAeMOFQHd50X8JO4dfZbmSo3YdJlVfz9FLRxE64mqj9bkN8hPFbkTG2F1AWXGPON5cmm4uiLPfQkWhX/LnClrhzZpNtMJYs5AEFeDs0POijcRugZsQA+wvLi0lSlhOfkqtjAJKpPUwy1wrJFDdvqdQBjpNQh/LB8c15XfQV2JT/3NX26dQe8zvHhL6NvfhBnAikodYkBr7UjSl36CBk0cPebZMZEBBiHdo76xORVkpmqDvkhFByXXeAsvRa2YWS4wxpiNJFswlRhjubGau7LrT113WMcPvgYXHYHf2IYJWD goneri@redhat.com"
-        state: present
-        user: zuul
-
     - name: reset ssh connection
       meta: reset_connection
 
--- a/playbooks/ansible-test-cloud-integration-base/files/bootstrap-vcsa.yaml
+++ b/playbooks/ansible-test-cloud-integration-base/files/bootstrap-vcsa.yaml
@@ -1,0 +1,33 @@
+---
+- hosts: appliance
+  gather_facts: false
+  tasks:
+    - name: Write the /etc/hosts file
+      copy:
+        content: |
+          127.0.0.1  localhost
+          ::1  localhost ipv6-localhost ipv6-loopback
+          {{ hostvars['vmware-vcsa-6.7.0']['nodepool']['interface_ip']}} vcenter.test vcenter
+        dest: /etc/hosts
+      become: true
+      register: hosts_status
+
+    - command: hostnamectl set-hostname vcenter.test
+      become: true
+
+    - name: Unconditionally reboot the machine with all defaults
+      reboot:
+      become: true
+      when: hosts_status.changed
+
+    - name: "wait for the vcenter service"
+      uri:
+        url: "https://127.0.0.1"
+        validate_certs: false
+        return_content: true
+      register: result
+      until:
+        - "result.status == 200"
+        - "'vmc-documentation-link' in result.content"
+      retries: 100
+      delay: 5
--- a/playbooks/ansible-test-cloud-integration-base/pre.yaml
+++ b/playbooks/ansible-test-cloud-integration-base/pre.yaml
@@ -1,0 +1,53 @@
+---
+- hosts: controller
+  tasks:
+    - name: Setup tox role
+      include_role:
+        name: tox
+      vars:
+        tox_extra_args: "-vv -- ansible-playbook -v playbooks/ansible-test-cloud-integration-base/files/bootstrap-vcsa.yaml"
+        tox_install_siblings: false
+        zuul_work_dir: "{{ zuul.projects['github.com/ansible/ansible-zuul-jobs'].src_dir }}"
+
+    - name: Setup base virtualenv_options
+      set_fact:
+        _virtualenv_options: "--python python{{ ansible_test_python }}"
+
+    - name: Enable --system-site-packages for virtualenv_options
+      set_fact:
+        _virtualenv_options: "{{ _virtualenv_options }} --system-site-packages"
+      when: ansible_os_family == "RedHat"
+
+    - name: Create virtualenv for ansible-test
+      shell: "virtualenv {{ _virtualenv_options }} ~/venv"
+
+    - name: Install ara into virtuelenv
+      shell: ~/venv/bin/pip install "ara<1.0.0"
+
+    - name: Upgrade six for vap-runtime (workaround)
+      shell: ~/venv/bin/pip install --upgrade "six>=1.12"
+
+
+    - name: Install ansible into virtuelenv
+      shell: "~/venv/bin/pip install {{ ansible_user_dir }}/{{ zuul.projects['github.com/ansible/ansible'].src_dir }}"
+
+    - name: Enable ARA callback plugin
+      ini_file:
+        path: "{{ ansible_user_dir }}/{{ zuul.projects['github.com/ansible/ansible'].src_dir }}/test/integration/integration.cfg"
+        section: defaults
+        option: callback_plugins
+        value: "{{ ansible_user_dir }}/venv/lib/python{{ ansible_test_python}}/site-packages/ara/plugins/callbacks"
+
+    - name: Enable persistent connection logging
+      ini_file:
+        path: "{{ ansible_user_dir }}/{{ zuul.projects['github.com/ansible/ansible'].src_dir }}/test/integration/integration.cfg"
+        section: persistent_connection
+        option: log_messages
+        value: true
+
+    - name: Set ansible log path
+      ini_file:
+        path: "{{ ansible_user_dir }}/{{ zuul.projects['github.com/ansible/ansible'].src_dir }}/test/integration/integration.cfg"
+        section: defaults
+        option: log_path
+        value: ~/ansible-debug.txt
--- a/playbooks/ansible-test-cloud-integration-base/pre.yaml
+++ b/playbooks/ansible-test-cloud-integration-base/pre.yaml
@@ -15,7 +15,7 @@
 
     - name: Enable --system-site-packages for virtualenv_options
       set_fact:
-        _virtualenv_options: "{{ _virtualenv_options }} --system-site-packages"
+        _virtualenv_options: "{{ _virtualenv_options }}"
       when: ansible_os_family == "RedHat"
 
     - name: Create virtualenv for ansible-test
@@ -23,10 +23,6 @@
 
     - name: Install ara into virtuelenv
       shell: ~/venv/bin/pip install "ara<1.0.0"
-
-    - name: Upgrade six for vap-runtime (workaround)
-      shell: ~/venv/bin/pip install --upgrade "six>=1.12"
-
 
     - name: Install ansible into virtuelenv
       shell: "~/venv/bin/pip install {{ ansible_user_dir }}/{{ zuul.projects['github.com/ansible/ansible'].src_dir }}"
--- a/playbooks/ansible-test-cloud-integration-base/run.yaml
+++ b/playbooks/ansible-test-cloud-integration-base/run.yaml
@@ -27,6 +27,6 @@
       args:
         chdir: "{{ _test_location }}"
         executable: /bin/bash
-      shell: "source ~/venv/bin/activate; {{ ansible_user_dir }}/{{ zuul.projects['github.com/ansible/ansible'].src_dir }}/bin/ansible-test integration {{ _test_options }} --python {{ ansible_test_python }} -vvvv {{ ansible_test_integration_target }}"
+      shell: "source ~/venv/bin/activate; {{ ansible_user_dir }}/{{ zuul.projects['github.com/ansible/ansible'].src_dir }}/bin/ansible-test integration {{ _test_options }} --python {{ ansible_test_python }} -vvvv {{ ansible_test_integration_targets }}"
       environment:
         VMWARE_TEST_PLATFORM: static
--- a/playbooks/ansible-test-cloud-integration-base/run.yaml
+++ b/playbooks/ansible-test-cloud-integration-base/run.yaml
@@ -1,0 +1,32 @@
+---
+- hosts: controller
+  tasks:
+    - name: Setup location of project
+      set_fact:
+        _test_location: "{{ ansible_user_dir }}/{{ zuul.projects['github.com/ansible/ansible'].src_dir }}"
+
+    - name: Setup base test_options
+      set_fact:
+        _test_options: "--continue-on-error --diff --requirements"
+
+    - name: Enable --no-temp-workdir for test_options
+      set_fact:
+        _test_options: "{{ _test_options }} --no-temp-workdir"
+      when: zuul.projects['github.com/ansible/ansible'].checkout not in ["stable-2.6", "stable-2.7"]
+
+    - copy:
+        content: |
+            [DEFAULT]
+            vcenter_username: administrator@vsphere.local
+            vcenter_password: !234AaAa56
+            vcenter_hostname: {{ hostvars['vmware-vcsa-6.7.0']['nodepool']['interface_ip']}}
+            vmware_validate_certs: false
+        dest: "{{ ansible_user_dir }}/{{ zuul.projects['github.com/ansible/ansible'].src_dir }}/test/integration/cloud-config-vcenter.ini"
+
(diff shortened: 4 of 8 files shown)


playbooks/ansible-cloud/appliance-base/pre.yaml
```diff
@@ -52,3 +52,10 @@
         write_inventory_dest: /etc/ansible/hosts
         write_inventory_exclude_hostvars:
           - ansible_connection
+
```
pabelanger: we should move this into ansible-test-cloud-integration-base bootstrap yaml file.  As these changes are specific to ansible-test.

That allows us to establish this job is for getting node online in zuul, and ansible-test-cloud-integration-base for things needed to make ansible-test work.

playbooks/ansible-test-cloud-integration-base/pre.yaml
```diff
@@ -0,0 +1,90 @@
...
+#    - name: Select proper ansible_network_os
+#      set_fact:
+#        _network_os: "{{ hostvars[groups['openvswitch'][0]]['ansible_network_os'] }}"
+#      when: '"openvswitch" in groups'
+#
+#    - name: Setup tox role
+#      include_role:
+#        name: tox
+#      vars:
+#        tox_extra_args: "-vv -- ansible-playbook -v playbooks/ansible-test-network-integration-base/files/bootstrap-{{ _network_os }}.yaml"
```
pabelanger: This is where the stuff above should move into. We run 2 nested bootstrap process for network, we should also do the same here.

playbooks/ansible-cloud/appliance-base/pre.yaml
goneri: ok, make sense.

playbooks/ansible-cloud/vsca-appliance/files/bootstrap.yaml
```diff
@@ -23,6 +23,35 @@
...
+        name: "zuul"
+        password: "{{ __password | password_hash('sha512') }}"
+      become: true
+
+    - name: Write the /etc/hosts file
+      copy:
+        content: |
+          127.0.0.1  localhost
+          ::1  localhost ipv6-localhost ipv6-loopback
+          {{ hostvars['vmware-vcsa-6.7.0']['nodepool']['private_ipv4']}} vcenter.test vcenter
```
pabelanger: you want ansible_host here, private_ip was only used for iosxr because of a bug in cisco networking.

playbooks/ansible-cloud/appliance-base/pre.yaml
```diff
@@ -52,3 +52,10 @@
...
+    - name: Add a vcenter entry in /etc/hosts
+      lineinfile:
+        dest: /etc/hosts
+        regexp: " vcenter "
+        line: "{{ hostvars['vmware-vcsa-6.7.0']['nodepool']['private_ipv4']}} vcenter.test vcenter"
```
pabelanger: same, ansible_host here. not private

playbooks/ansible-cloud/vsca-appliance/files/bootstrap.yaml
```diff
@@ -23,6 +23,35 @@
...
+        dest: /etc/hosts
+      become: true
+      register: hosts_status
+
+    - name: Unconditionally reboot the machine with all defaults
```
pabelanger: I suspect this is going to be really expensive to do, we any info on what you are trying to do?

goneri:
recheck

playbooks/ansible-cloud/vsca-appliance/files/bootstrap.yaml
goneri: It's a hack to restart all the services with the correct reverse DNS. The reboot takes 30s.
This will go away the day I find a better solution.

playbooks/ansible-test-cloud-integration-base/pre.yaml
```diff
@@ -0,0 +1,90 @@
...
+      shell: "virtualenv {{ _virtualenv_options }} ~/venv"
+
+    - name: Install ara into virtuelenv
+      shell: ~/venv/bin/pip install "ara<1.0.0"
+
+    - name: Install ansible into virtuelenv
+      shell: "~/venv/bin/pip install {{ ansible_user_dir }}/{{ zuul.projects['github.com/ansible/ansible'].src_dir }}"
+
+    - name: Upgrade setuptools into virtualenv
+      shell: "~/venv/bin/pip install --upgrade pip setuptools 'six>=1.12'"
```
pabelanger: I'd be surprised if we need to update setuptools, latest will be already installed.  six should go into the requirements file in ansible/ansible, then tihs PR should use depends-on, to pull it in t for testing.

playbooks/ansible-test-cloud-integration-base/pre.yaml
```diff
@@ -0,0 +1,90 @@
...
+
+    - name: Install vSphere Automation SDK
+      shell: "~/venv/bin/pip install --upgrade git+https://github.com/vmware/vsphere-automation-sdk-python.git"
+
+    - name: Install pyvmomi and jmespath
+      shell: "~/venv/bin/pip install --upgrade pyvmomi jmespath"
```
pabelanger: same comment, these should be handled in ansible./ansible requirements, we should have the latest versions

playbooks/ansible-cloud/appliance-base/pre.yaml
goneri: ok. Should be addressed in the next update.

playbooks/ansible-cloud/vcsa-appliance/files/bootstrap.yaml
```diff
@@ -1,4 +1,52 @@
...
+    - name: Enable --system-site-packages for virtualenv_options
+      set_fact:
+        _virtualenv_options: "{{ _virtualenv_options }} --system-site-packages"
+      when: ansible_os_family == "RedHat"
+
+    - name: Create virtualenv for ansible-test
+      shell: "virtualenv {{ _virtualenv_options }} ~/venv"
+
+    - name: Install ansible into virtualenv
+      shell: "~/venv/bin/pip install {{ ansible_user_dir }}/{{ zuul.projects['github.com/ansible/ansible'].src_dir }} -rtest/lib/ansible_test/_data/requirements/integration.cloud.vcenter.txt jmespath"
```
pabelanger: we don't need to do this, as ansible-test will manage its own dependencies with the --requirements flag.

playbooks/ansible-cloud/vcsa-appliance/files/bootstrap.yaml
```diff
@@ -1,4 +1,52 @@
 ---
+- hosts: controller
+  tasks:
+    - name: Add a vcenter entry in /etc/hosts
```
pabelanger: we can move this into files/bootstrap.yaml, which will allow us to abstract this playbook out into shared content between network /cloud jobs.

playbooks/ansible-test-cloud-integration-base/run.yaml
```diff
@@ -0,0 +1,101 @@
...
+#        - name: Setup location of project
+#          set_fact:
+#            _test_location: "~/.ansible/collection/ansible_collections/{{ _collection_namespace.stdout }}/{{ _collection_name.stdout }}"
+#
+#        - name: network-integration.cfg
+#          shell: "cp {{ ansible_user_dir }}/{{ zuul.projects['github.com/ansible/ansible'].src_dir }}/test/integration/network-integration.cfg {{ _test_location }}/tests/integration/network-integration.cfg"
+#
+#      when: ansible_test_collections | default(False)
+
+    - copy:
```
pabelanger: Lets switch this to template, so we can change variables to be dynamic.

playbooks/ansible-test-cloud-integration-base/run.yaml
```diff
@@ -0,0 +1,101 @@
+---
```
pabelanger: I image we'll make this playbook generic enough to support both jobs, rather then copypasta this whole thing

playbooks/ansible-cloud/appliance-base/pre.yaml
```diff
@@ -1,54 +1,82 @@
+# TODO DUP of ansible-test-network-integration-base/pre.yaml
```
pabelanger: This is the wrong file, the file that will be dupe will be

playbooks/ansible-test-cloud-integration-base/pre.yaml 

We don't need to modify anything here.

playbooks/ansible-test-cloud-integration-base/files/bootstrap-vcsa.yaml
```diff
@@ -0,0 +1,35 @@
+- hosts: appliance
+  gather_facts: false
+  vars:
+    ansible_password: zuul
```
pabelanger: this isn't needed, we'll use SSH public keys to connection now

playbooks/ansible-test-cloud-integration-base/files/bootstrap-vcsa.yaml
```diff
@@ -0,0 +1,35 @@
...
+    ansible_user: zuul
```
pabelanger: same

playbooks/ansible-test-cloud-integration-base/pre.yaml
```diff
@@ -0,0 +1,53 @@
...
+        zuul_work_dir: "{{ zuul.projects['github.com/ansible/ansible-zuul-jobs'].src_dir }}"
+
+    - name: Setup base virtualenv_options
+      set_fact:
+        _virtualenv_options: "--python python{{ ansible_test_python }}"
+
+    - name: Enable --system-site-packages for virtualenv_options
+      set_fact:
+        _virtualenv_options: "{{ _virtualenv_options }} --system-site-packages"
+      when: ansible_os_family == "RedHat"
```
pabelanger: I believe this is the issue for six, as we are using the system version inside the venv. My fix for this, is to drop --system-site-packages from virtualenv, and use https://pypi.org/project/selinux/

I tested it before and worked, I can open another PR to confirm network jobs don't break using it.

playbooks/ansible-test-cloud-integration-base/pre.yaml
```diff
@@ -0,0 +1,53 @@
...
+      shell: ~/venv/bin/pip install "ara<1.0.0"
+
+    - name: Upgrade six for vap-runtime (workaround)
+      shell: ~/venv/bin/pip install --upgrade "six>=1.12"
+
+
+    - name: Install ansible into virtuelenv
+      shell: "~/venv/bin/pip install {{ ansible_user_dir }}/{{ zuul.projects['github.com/ansible/ansible'].src_dir }}"
+
+    - name: Enable ARA callback plugin
```
pabelanger: okay, below won't work with integration, so we likely need to modify another cfg file.

pabelanger: okay, lets change to test/lib/ansible_test/_data/ansible.cfg for now

playbooks/ansible-test-cloud-integration-base/files/bootstrap-vcsa.yaml
goneri: I'm actually surprised it works. The password should have been changed.

playbooks/ansible-test-cloud-integration-base/pre.yaml
pabelanger: test/integration/integration.cfg

that is what we need to create

playbooks/ansible-test-cloud-integration-base/pre.yaml
goneri: Indeed. I will let the six thing aside for now. It will be addressed in another PR.

playbooks/ansible-test-cloud-integration-base/pre.yaml
goneri: +1, nice catch.

playbooks/ansible-test-cloud-integration-base/files/bootstrap-vcsa.yaml
pabelanger: so, 3 things I can think of. 

1) we need to now disable ssh access using passwords, this is in sshd.conf
2) maybe we didn't save the config for passwords.
3) maybe it isn't even working, as ssh public keys get priority over password

playbooks/ansible-test-cloud-integration-base/run.yaml
goneri: Agreed.

playbooks/ansible-test-cloud-integration-base/run.yaml
goneri: It's an ini file, we can use the `ini_file` module instead.

playbooks/ansible-test-cloud-integration-base/files/bootstrap-vcsa.yaml
goneri: or SSH keeps using the same controlpath.

pabelanger:
for reasons, depends-on has to be first comment in github PR. I want to change that.

pabelanger:
recheck

zuul.d/project-templates.yaml
```diff
@@ -1149,6 +1149,131 @@
               - ^lib/ansible/plugins/terminal/vyos.py
               - ^test/integration/targets/vyos_.*
               - ^test/integration/targets/prepare_vyos_tests/.*
+        - ansible-test-cloud-integration-vcsa-python36:
+            branches:
+              - devel
+            files:
```
pabelanger: can we simplify this?

Maybe something like

^lib/ansible/modules/cloud/vmware/*
^lib/ansible/module_utils/vmware.*
 ^test/integration/targets/vcenter.*

to start

goneri: Just a subset of the vmware modules can be tested with just one vcenter. So I'm not sure it makes sense to trigger the test if we change a module that is not covered. It sounds like just a waste of resources.

playbooks/ansible-test-cloud-integration-base/files/bootstrap-vcsa.yaml
```diff
@@ -0,0 +1,33 @@
...
+    - name: "wait for the vcenter service"
+      uri:
+        url: "https://127.0.0.1"
+        validate_certs: false
+        return_content: true
+      register: result
+      until:
+        - "result.status == 200"
+        - "'vmc-documentation-link' in result.content"
+      retries: 18000
```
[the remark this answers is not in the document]
goneri: ahah, I forgot this one. Indeed we can lower it.

zuul.d/project-templates.yaml
pabelanger: I wouldn't say a waste of resources, once we move to collections, we'll be running this job all the time.  But yes, any changes to vmware, I'd say should trigger our vcsa job for now. There is a trade off, for me history has taught me, people tend to get confused if their jobs doesn't run, if the matcher gets out of sync with new files, etc. For that reason along, tend to make the matcher as large as possible.

But even here, ^lib/ansible/module_utils/vmware.* is duplicated a bunch of times, this isn't need. And possible a copypaste issue?

goneri: Ok I will trigger the job for all the vmware changes. And indeed, the dup ^lib/ansible/module_utils/vmware.* comes from a bug in my script. Let's fix that.

playbooks/ansible-cloud/vcsa-appliance/files/bootstrap.yaml
```diff
@@ -15,9 +15,12 @@
         state: present
         user: zuul
 
-- hosts: appliance
```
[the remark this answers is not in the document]
goneri: Done, I aslo added a `reset_connection` call to be sure we don't controlpath socket opened.

playbooks/ansible-test-cloud-integration-base/pre.yaml
goneri: Ok, it works fine without the `--system-site-packages`, thanks for the idea.

playbooks/ansible-test-cloud-integration-base/run.yaml
```diff
@@ -0,0 +1,32 @@
...
+            vcenter_password: !234AaAa56
+            vcenter_hostname: {{ hostvars['vmware-vcsa-6.7.0']['nodepool']['interface_ip']}}
+            vmware_validate_certs: false
+        dest: "{{ ansible_user_dir }}/{{ zuul.projects['github.com/ansible/ansible'].src_dir }}/test/integration/cloud-config-vcenter.ini"
+
+    - name: Run the integration test suite
+      args:
+        chdir: "{{ _test_location }}"
+        executable: /bin/bash
+      shell: "source ~/venv/bin/activate; {{ ansible_user_dir }}/{{ zuul.projects['github.com/ansible/ansible'].src_dir }}/bin/ansible-test integration {{ _test_options }} --python {{ ansible_test_python }} -vvvv {{ ansible_test_integration_target }}"
```
pabelanger: nit, ansible_test_integration_targets, as we can pass multiple

zuul.d/ansible-cloud-jobs.yaml
```diff
@@ -15,3 +15,23 @@
...
+    timeout: 3600
+
+- job:
+    name: ansible-test-cloud-integration-vcsa-python36
+    parent: ansible-test-cloud-integration-vcsa
+    nodeset: vmware-vcsa-6.7.0-python36
+    vars:
+      ansible_test_python: 3.6
+      ansible_test_integration_target: zuul/vmware/vcenter_only/
+    attempts: 1
```
pabelanger: remember to remove this once everything is 100%

zuul.d/project.yaml
```diff
@@ -13,6 +13,10 @@
             voting: false
         - ansible-network-openvswitch-appliance
         - ansible-network-vyos-appliance
+        - ansible-test-cloud-integration-vcsa-python36:
```
pabelanger: remember to remove and test from ansible/ansible

playbooks/ansible-test-cloud-integration-base/run.yaml
```diff
@@ -0,0 +1,32 @@
...
+      environment:
+        VMWARE_TEST_PLATFORM: static
```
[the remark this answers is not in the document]
goneri: What?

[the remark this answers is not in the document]
pabelanger: sorry,

VMWARE_TEST_PLATFORM: static

First time I have seen environmental var for ansible-test

playbooks/ansible-test-cloud-integration-base/pre.yaml
goneri: Hum, Why should I? I don't need to pin on a specific six version anymore.

playbooks/ansible-test-cloud-integration-base/run.yaml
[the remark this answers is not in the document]
goneri: The vcenter provider can work with tree different platforms:

- vcsim
- static (an existing platform)
- worldstream, in this case it uses the web API to start a remote environment first.

The environment allow us to specify which one we need.

playbooks/ansible-test-cloud-integration-base/pre.yaml
pabelanger: This was originally dealing with selinux errors controller side, when updating things like file permissions. Given that it works now, makes me think, this isn't needed any more on centos-8 but is something we could hit in fedora.

Lets leave this for now, the longer term fix is to drop this, but I think we also need selinux package for selinux python bindings inside the venv.  I'll work on testing that this morning.

zuul.d/ansible-cloud-jobs.yaml
goneri: Done, this being said, the default '3' is totally overkill IMO.

pabelanger: Agree in perfect world, this really protects on something funky happening on the network, which we do see in both vexxhost / limestone.  With out it, we'll see RETRY_LIMIT failures on jobs.... which we do still, just much lower rate as we try 3 times for jobs to run.

playbooks/ansible-test-cloud-integration-base/run.yaml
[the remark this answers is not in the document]
pabelanger: thanks!

zuul.d/project.yaml
```diff
@@ -13,6 +13,7 @@
             voting: false
         - ansible-network-openvswitch-appliance
         - ansible-network-vyos-appliance
+
```
pabelanger: whitespace, can be removed

playbooks/ansible-cloud/vcsa-appliance/files/bootstrap.yaml
```diff
@@ -15,9 +15,19 @@
...
+        state: present
+        user: zuul
+
+    - name: reset ssh connection
+      meta: reset_connection
+
 - hosts: appliance
   gather_facts: false
   tasks:
+
```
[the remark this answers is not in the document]
pabelanger: whitespace

PR #255, opened 2019-11-28
Update ios-15.6-2T-python36 to centos-8

pabelanger:
With centos-8 online, now we can use it for python36 support.


pabelanger:
recheck

PR #263, opened 2019-12-06
speed up the vcenter tests

goneri:
Skip the reboot step, it's not a requirement anymore
    
Two last runs without the patch took 2167s and 2311s. First run with the
patch takes 1757s. We save =~ 8m per run.


playbooks/ansible-cloud/vcsa-appliance/files/bootstrap.yaml
```diff
@@ -1,34 +1,4 @@
 ---
 - hosts: appliance
   gather_facts: false
-  vars:
```
pabelanger: we don't want to delete this, as new jobs will parent to ansible-cloud-vsca-appliance directly. So SSH key falls into still.

playbooks/ansible-cloud/vcsa-appliance/pre.yaml
```diff
@@ -8,10 +8,52 @@
         name: sshpass
         state: present
 
-    - name: Setup tox role
```
pabelanger: This will not work, as zuul doesn't have have SSH info to connect to this VM, this is why we still need embedded bootstrap.yaml files.

playbooks/ansible-test-cloud-integration-base/pre.yaml
```diff
@@ -1,14 +1,6 @@
 ---
 - hosts: controller
   tasks:
-    - name: Setup tox role
```
pabelanger: We still need this

playbooks/ansible-test-cloud-integration-base/run.yaml
```diff
@@ -14,6 +14,18 @@
         _test_options: "{{ _test_options }} --no-temp-workdir"
       when: zuul.projects['github.com/ansible/ansible'].checkout not in ["stable-2.6", "stable-2.7"]
 
+    - name: "wait for the vcenter service"
```
pabelanger: this runs from zuul-executor, which cannot connect to the appliance directly.

pabelanger: actually, this is something we should test. Maybe zuul-executor will have correct SSH info, after we create it.

goneri: It's an HTTP call.

playbooks/ansible-cloud/vcsa-appliance/files/bootstrap.yaml
goneri: I don't want to delete, I just would like to move it out of the nested Ansible. We prepare a tox env just to run a couple of tasks. This is a bit overkill.

Anyway, I reverted this part in my next iteration.

pabelanger: Yes, it is overkill, but a known pattern right now with network jobs. I don't mind we refactor, we need to do all the same.

IMO, we'll get better improvement if we don't run prepare_vmware_tasks multiple times, that is cumulative cycles, for each playbook we run.

playbooks/ansible-test-cloud-integration-base/pre.yaml
pabelanger: So, I'd like to keep this for now, for our http check in vmware. As, the next step I would like to do, is refactor

playbooks/ansible-test-cloud-integration-base/run.yaml
playbooks/ansible-test-network-integration-base/run.yaml 

into a shared playbook.

While I agree, this change seems to make sense here, it will not in network. As network cannot talk to appliances today.

goneri:
> I think I;d just like to remove the reboot stuff right now, and not refactor tox bit. Until we can refactor network and cloud to use single playbook for ansible-test.

I reverted half of the patch, but I still remove `playbooks/ansible-test-cloud-integration-base/files/bootstrap-vcsa.yaml`. This because beside the HTTP test, nothing was done there anymore. And it's actually better to do the HTTP check from another node BTW.

pabelanger:
Have we tested this via depends-on in ansible/ansible? If so, can you link here to make easier to find? This looks correct from what I see.

goneri:
@pabelanger  https://object-storage-ca-ymq-1.vexxhost.net/v1/a0b4156a37f9453eb4ec7db5422272df/ansible_16/65616/18c898dde6682ae96020fce9667b4ddf1d3108c5/third-party-check/ansible-test-cloud-integration-vcsa-python36/b1c26e4/

PR #270, opened 2019-12-10
rename vcsa to vcenter for clarity

goneri:
VCSA stands for vCenter Server Appliance, and so is actually a vcenter.
vCenter is a much more popular terminology. Let's avoid any confusion
and just use just this therm.

In addition, we are about to introduce more VMware jobs, and the number
of nodes will be the distinguish factor. So when it makes sense, this
commit uses 1vcenter for clarity.


zuul.d/nodesets.yaml
```diff
@@ -31,16 +31,16 @@
 
 # Ansible cloud appliance nodesets
 - nodeset:
-    name: vmware-vcsa-6.7.0-python36
```
pabelanger: I don't think we can do this, because we want to support multiple versions of vsca for stable branches.

goneri: As I explain in the description, the vcsa term just brings confusion without any real value. So I would still prefer something like `vmware-vcenter-6.7.0-python36`.

When the ESXi test will be in place, the name will be seriously over complicated. e.g: 
vmware-1vcenter-6.7.0U3b-2esxi-6.7.0U3-python36

I would prefer we keep rather agnostic nodeset names. Also do we really need the Python version there?:

- vmware-1vcenter-2esxi-6.7
- vmware-1vcenter-1esxi-6.5

pabelanger: I think we can rename the job, but the nodeset maybe not. Otherwise, we have no way to bring online a different version, which is something we need to do to support stable branches still.

So the job name is fine, I think. We also want python36, as we plan on adding python27, python35, python37 and python38 jobs.  In zuul, each job still needs a unique name.

goneri: > I think we can rename the job, but the nodeset maybe not. Otherwise, we have no way to bring online a different version, which is something we need to do to support stable branches still.

:thinking: We can still us a suffix, like `lastest` for 6.7.0U3, `legacy` for 6.5. This being said, we may want to mix several versions at some point:

- vmware-vcenter-only-latest
- vmware-vcenter-only-legacy
- vmware-vcenter-1esxi-legacy

> So the job name is fine, I think. We also want python36, as we plan on adding python27, python35, python37 and python38 jobs. In zuul, each job still needs a unique name.

Do we really need the python version in the nodeset. The version depends on the controller, and this node may provider some other Python versions.

pabelanger: Well, job name is fine, we have the ability to create variants in zuul. So we don't need latest or 6.5 in the name, just the nodeset.

The python version in question, is controller side. For that we need to support 5 python versions, this is what we do today in network

pabelanger: So, for a single branch, I'd expect us to only run a single version of vcenter. And for older branches, maybe another. The job name is the same, but we'd have a different nodeset. This is called a job variant in zuul.

playbooks/ansible-cloud/vcenter-appliance/files/bootstrap.yaml
```diff
@@ -33,10 +33,10 @@
         name: zuul
         password: "{{ __password | password_hash('sha512') }}"
 
-- hosts: vmware-vcsa-6.7.0
+- hosts: vcenter
```
pabelanger: We likely want this to be the group name, so appliance, right? As we could have more then 1 vcenter node.

goneri: I think it's fine since we will have both esxi and vcenter in appliance group. Here we only target the vcenter. And so far we don't have any test with two vcenter.

goneri:
@pabelanger does this look better? Follow up to you discussion, I'm not sure if I can/cannot remove the python version like I did.

zuul.d/nodesets.yaml
```diff
@@ -35,12 +35,12 @@
     nodes:
       - name: centos-8
         label: centos-8-1vcpu
-      - name: vmware-vcsa-6.7.0
+      - name: vcenter
```
pabelanger: I still think it is helpful to have version number listed in console logs or inventory, but we can loop back once we add a 2nd version.

zuul.d/project.yaml
```diff
@@ -16,6 +16,7 @@
         - ansible-network-vyos-appliance
     gate:
       jobs:
+        - ansible-cloud-vcenter-appliance
```
pabelanger: This needs to be removed, we flipped the job non-voting because of timeout failure we seen from zuul-executor

zuul.d/nodesets.yaml
goneri: yes, but no. Sorry. The vcenter hostname during the test is `vcenter.test` anyway, and we cannot change it. In addition, I want the whole automation to be agnostic as possible.

zuul.d/project.yaml
goneri: oh, ok let's me fix that

goneri: Done

pabelanger:
recheck

PR #271, opened 2019-12-11
vmware: add a vcenter+1esxi job

goneri:
This rather large commit introduce a new vcenter job. This time, we have
an ESXi instance and this allow us to run most of the missing test
roles.

- We use the `zuul` of the ESXi instead of the classic `root`. This
  because the root account get hammered by SSH connection attempts, and
  ultimately... locked (in a couple seconds actually :-P).
- Some weird gymnastic around the vcenter password. vCenter generates it
  admin password itself, we cannot set one that we've generated before.
  The logic is wrapped in a role, this way we can reproduce/test out of
  Zuul.
- host names are enforced via /etc/hosts. vCenter is pathological picky
  with host reverse DNS. By changing the /etc/hosts, we can keep
  everything under control and reproductible. Here again, the logic is
  in a role.
- NFS server is also deployed using a role, to improve the resilience we
  should set-up some cache somewhere for the ISO images.
- The test-suite itself is known to fail. So `attempt` will remain at 1
  for now.

### DIFF
--- a/playbooks/ansible-cloud/appliance-base/pre.yaml
+++ b/playbooks/ansible-cloud/appliance-base/pre.yaml
@@ -11,19 +11,9 @@
       include_role:
         name: test-setup
 
-    - name: Set SSH interface IP for appliance
-      set_fact:
-        __ssh_ip: "{{ hostvars[item].ansible_host }}"
-      with_inventory_hostnames: appliance
-
-    - name: Wait 300 seconds for port tcp/22 on appliance
-      wait_for:
-        host: "{{ __ssh_ip }}"
-        port: 22
-        search_regex: SSH
-
     - name: Ensure remote SSH host keys are known
-      shell: "ssh-keyscan {{ __ssh_ip }} >> ~/.ssh/known_hosts"
+      shell: "ssh-keyscan {{hostvars[item].ansible_host }} >> ~/.ssh/known_hosts"
+      with_inventory_hostnames: appliance
 
     - name: Create /etc/ansible folder
       become: true
--- a/playbooks/ansible-cloud/appliance-base/pre.yaml
+++ b/playbooks/ansible-cloud/appliance-base/pre.yaml
@@ -11,6 +11,17 @@
       include_role:
         name: test-setup
 
+    - name: Set SSH interface IP for appliance
+      set_fact:
+        __ssh_ip: "{{ hostvars[item].ansible_host }}"
+      with_inventory_hostnames: appliance
+
+    - name: Wait 300 seconds for port tcp/22 on appliance
+      wait_for:
+        host: "{{ __ssh_ip }}"
+        port: 22
+        search_regex: SSH
+
     - name: Ensure remote SSH host keys are known
       shell: "ssh-keyscan {{hostvars[item].ansible_host }} >> ~/.ssh/known_hosts"
       with_inventory_hostnames: appliance
--- a/playbooks/ansible-cloud/vcenter-appliance/files/bootstrap.yaml
+++ b/playbooks/ansible-cloud/vcenter-appliance/files/bootstrap.yaml
@@ -4,7 +4,12 @@
   vars:
     ansible_password: zuul
     ansible_user: zuul
+    authorized_keys_file:
+      esxi1: /etc/ssh/keys-zuul/authorized_keys
+      esxi2: /etc/ssh/keys-zuul/authorized_keys
+      vcenter: /home/zuul/.ssh/authorized_keys
   tasks:
+
     - name: lookup SSH public key
       set_fact:
         _ssh_key: "{{ lookup('file', '~/.ssh/id_rsa.pub') }}"
@@ -14,6 +19,7 @@
         key: "{{ _ssh_key }}"
         state: present
         user: zuul
+        path: "{{ authorized_keys_file[inventory_hostname] }}"
 
     - name: reset ssh connection
       meta: reset_connection
@@ -24,29 +30,17 @@
 
     - name: Generate random password
       set_fact:
-        __password: "{{ lookup('password', '/dev/null length=15 chars=ascii_letters') }}"
+        __password: "{{ lookup('password', '/dev/null length=15 chars=ascii_letters,digits,hexdigits') }}"
       no_log: true
 
-    - name: Change zuul user password
+    - name: Change zuul user password (vcenter)
       become: true
       user:
         name: zuul
         password: "{{ __password | password_hash('sha512') }}"
+      when: inventory_hostname == "vcenter"
 
-- hosts: vcenter
-  gather_facts: false
-  tasks:
-    # NOTE(Gonéri): ok, some explanations here. We can set the hostname of the vcenter node
-    # during the installation, and we can hardly change it after. So we configure the VM
-    # to use dnsmasq internally and use /etc/hosts as a source of trust.
-    # Also, the hostname has to be in the format $hostname.$domain. In addition, some services
-    # use the reverse DNS resolution to know the public IP of the node. So we cannot just go
-    # with "vcenter.test==127.0.0.1", this won't work.
-    - name: Write the /etc/hosts file
-      copy:
-        content: |
-          127.0.0.1  localhost
-          ::1  localhost ipv6-localhost ipv6-loopback
-          {{ hostvars['vcenter']['nodepool']['interface_ip']}} vcenter.test vcenter
-        dest: /etc/hosts
-      become: true
+    - name: Change zuul user password (ESXi)
+      shell: "echo {{ __password }} | passwd -s"
+      no_log: true
+      when: inventory_hostname.startswith("esxi")
--- a/playbooks/ansible-cloud/vcenter-appliance/files/bootstrap.yaml
+++ b/playbooks/ansible-cloud/vcenter-appliance/files/bootstrap.yaml
@@ -23,24 +23,3 @@
 
     - name: reset ssh connection
       meta: reset_connection
-
-- hosts: appliance
-  gather_facts: false
-  tasks:
-
-    - name: Generate random password
-      set_fact:
-        __password: "{{ lookup('password', '/dev/null length=15 chars=ascii_letters,digits,hexdigits') }}"
-      no_log: true
-
-    - name: Change zuul user password (vcenter)
-      become: true
-      user:
-        name: zuul
-        password: "{{ __password | password_hash('sha512') }}"
-      when: inventory_hostname == "vcenter"
-
-    - name: Change zuul user password (ESXi)
-      shell: "echo {{ __password }} | passwd -s"
-      no_log: true
-      when: inventory_hostname.startswith("esxi")
--- a/playbooks/ansible-test-cloud-integration-base/pre.yaml
+++ b/playbooks/ansible-test-cloud-integration-base/pre.yaml
@@ -12,6 +12,33 @@
         tox_install_siblings: false
         zuul_work_dir: "{{ zuul.projects['github.com/ansible/ansible-zuul-jobs'].src_dir }}"
 
+
+    - name: Restore the cache files for the NFS datastore
+      block:
+        - name: Create the ISO directory
+          file:
+            path: /srv/share/isos
+            state: directory
+          become: true
+
+        - name: Move the cached files
+          shell: |
+             mv /opt/cache/files/CentOS-8-x86_64-1905-dvd1.iso /srv/share/isos/centos.iso
+             mv /opt/cache/files/Fedora-Workstation-Live-x86_64-30-1.2.iso /srv/share/isos/fedora.iso
+          args:
+            creates: /srv/share/isos/fedora.iso
+          become: true
+
+        - name: Adjust the ownership
+          file:
+            path: /srv/share/isos
+            state: directory
+            recurse: true
+            owner: '1000'
+            group: '1000'
+          become: true
+
+
     - name: Prepare the NFS datastore
       import_role:
         name: vmware-ci-nfs-share
--- a/playbooks/ansible-test-cloud-integration-base/pre.yaml
+++ b/playbooks/ansible-test-cloud-integration-base/pre.yaml
@@ -23,8 +23,8 @@
 
         - name: Move the cached files
           shell: |
-             mv /opt/cache/files/CentOS-8-x86_64-1905-dvd1.iso /srv/share/isos/centos.iso
-             mv /opt/cache/files/Fedora-Workstation-Live-x86_64-30-1.2.iso /srv/share/isos/fedora.iso
+             mv /opt/cache/files/CentOS-8-x86_64-1905-boot.iso /srv/share/isos/centos.iso
+             mv /opt/cache/files/Fedora-Workstation-Live-x86_64-31-1.9.iso /srv/share/isos/fedora.iso
           args:
             creates: /srv/share/isos/fedora.iso
           become: true
--- a/playbooks/ansible-test-cloud-integration-base/pre.yaml
+++ b/playbooks/ansible-test-cloud-integration-base/pre.yaml
@@ -1,6 +1,9 @@
 ---
+
 - hosts: controller
+  gather_facts: false
   tasks:
+
     - name: Setup tox role
       include_role:
         name: tox
@@ -9,11 +12,27 @@
         tox_install_siblings: false
         zuul_work_dir: "{{ zuul.projects['github.com/ansible/ansible-zuul-jobs'].src_dir }}"
 
-    - copy:
-        content: |
-            [DEFAULT]
-            vcenter_username: administrator@vsphere.local
-            vcenter_password: !234AaAa56
-            vcenter_hostname: {{ hostvars['vcenter']['nodepool']['interface_ip']}}
-            vmware_validate_certs: false
-        dest: "{{ ansible_user_dir }}/{{ zuul.projects['github.com/ansible/ansible'].src_dir }}/test/integration/cloud-config-vcenter.ini"
+    - import_role:
+        name: vmware-ci-nfs-share
+      vars:
+        vmware_ci_nfs_share_allow_ips:
+          - "{{ hostvars['esxi1']['nodepool']['interface_ip'] }}"
+      when: "'esxi1' in hostvars"
+
+    - import_role:
+        name: vmware-ci-configure-ansible-test
+      vars:
+        vmware_ci_configure_ansible_test_ansible_path: "{{ ansible_user_dir }}/{{ zuul.projects['github.com/ansible/ansible'].src_dir }}"
+        vmware_ci_set_passwords_secret_dir: '{{ zuul.executor.work_root }}'
+
+    - name: "wait for the vcenter service"
+      uri:
+        url: "https://vcenter.test"
+        validate_certs: false
+        return_content: true
+      register: result
+      until:
+        - "result.status == 200"
+        - "'vmc-documentation-link' in result.content"
+      retries: 100
+      delay: 5
--- a/playbooks/ansible-test-cloud-integration-base/pre.yaml
+++ b/playbooks/ansible-test-cloud-integration-base/pre.yaml
@@ -12,20 +12,22 @@
         tox_install_siblings: false
         zuul_work_dir: "{{ zuul.projects['github.com/ansible/ansible-zuul-jobs'].src_dir }}"
 
-    - import_role:
+    - name: Prepare the NFS datastore
+      import_role:
         name: vmware-ci-nfs-share
       vars:
         vmware_ci_nfs_share_allow_ips:
           - "{{ hostvars['esxi1']['nodepool']['interface_ip'] }}"
       when: "'esxi1' in hostvars"
 
-    - import_role:
+    - name: Write the configuration for ansible-test
+      import_role:
         name: vmware-ci-configure-ansible-test
       vars:
         vmware_ci_configure_ansible_test_ansible_path: "{{ ansible_user_dir }}/{{ zuul.projects['github.com/ansible/ansible'].src_dir }}"
         vmware_ci_set_passwords_secret_dir: '{{ zuul.executor.work_root }}'
 
-    - name: "wait for the vcenter service"
+    - name: wait for the vcenter service
       uri:
         url: "https://vcenter.test"
         validate_certs: false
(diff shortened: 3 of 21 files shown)


goneri:
recheck

playbooks/ansible-cloud/vcenter-appliance/files/bootstrap.yaml
```diff
@@ -0,0 +1,77 @@
+---
+- hosts: appliance
+  gather_facts: false
+  vars:
+    ansible_ssh_extra_args: '-o StrictHostKeyChecking=no'
```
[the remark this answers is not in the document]
goneri: This branch is full of hack/work around. It's not really ready for prime time.

zuul.d/nodesets.yaml
```diff
@@ -31,16 +31,35 @@
 
 # Ansible cloud appliance nodesets
 - nodeset:
-    name: vmware-vcsa-6.7.0-python36
+    name: vmware-vcsa-6.7.0
     nodes:
       - name: centos-8
         label: centos-8-1vcpu
-      - name: vmware-vcsa-6.7.0
+      - name: vcenter
```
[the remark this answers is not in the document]
goneri: We don't have any test with two vcenter and the information is already defined in `hostvars['vcenter']['nodepool']['label']`.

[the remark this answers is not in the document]
goneri: I also feel more comfortable doing `hostvars['vcenter']['nodepool']['interface_ip']`, instead of going through a group.

goneri:
recheck

goneri:
recheck

goneri:
recheck

goneri:
recheck

goneri:
recheck

goneri:
recheck

goneri:
recheck

goneri:
recheck

playbooks/ansible-cloud/appliance-base/pre.yaml
```diff
@@ -11,19 +11,9 @@
...
-      set_fact:
-        __ssh_ip: "{{ hostvars[item].ansible_host }}"
-      with_inventory_hostnames: appliance
-
-    - name: Wait 300 seconds for port tcp/22 on appliance
-      wait_for:
-        host: "{{ __ssh_ip }}"
-        port: 22
-        search_regex: SSH
-
```
pabelanger: We actually want to keep this logic, as we don't have a network check from zuul-executor. What this protects against, is if the appliance doesn't come online properly, we abort the job and reschedule it.

playbooks/ansible-cloud/vcenter-appliance/files/bootstrap.yaml
```diff
@@ -0,0 +1,53 @@
...
+        state: present
+        user: zuul
+        path: "{{ authorized_keys_file[inventory_hostname] }}"
+
+    - name: Setup zuul user
+      authorized_key:
+        key: "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQCyiXfE1zHKdj6odbysr917Cn88ov0VQaPJtLKJyMNuRYAeMOFQHd50X8JO4dfZbmSo3YdJlVfz9FLRxE64mqj9bkN8hPFbkTG2F1AWXGPON5cmm4uiLPfQkWhX/LnClrhzZpNtMJYs5AEFeDs0POijcRugZsQA+wvLi0lSlhOfkqtjAJKpPUwy1wrJFDdvqdQBjpNQh/LB8c15XfQV2JT/3NX26dQe8zvHhL6NvfhBnAikodYkBr7UjSl36CBk0cPebZMZEBBiHdo76xORVkpmqDvkhFByXXeAsvRa2YWS4wxpiNJFswlRhjubGau7LrT113WMcPvgYXHYHf2IYJWD goneri.lebouder@enovance.com"
+        state: present
+        user: zuul
+        path: "{{ authorized_keys_file[inventory_hostname] }}"
```
pabelanger: will need to be deleted, added when we create autoholds

playbooks/ansible-test-cloud-integration-base/pre.yaml
```diff
@@ -1,19 +1,44 @@
...
-            vcenter_username: administrator@vsphere.local
-            vcenter_password: !234AaAa56
-            vcenter_hostname: {{ hostvars['vmware-vcsa-6.7.0']['nodepool']['interface_ip']}}
-            vmware_validate_certs: false
-        dest: "{{ ansible_user_dir }}/{{ zuul.projects['github.com/ansible/ansible'].src_dir }}/test/integration/cloud-config-vcenter.ini"
+    - name: Setup zuul user
+      authorized_key:
+        key: "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQCyiXfE1zHKdj6odbysr917Cn88ov0VQaPJtLKJyMNuRYAeMOFQHd50X8JO4dfZbmSo3YdJlVfz9FLRxE64mqj9bkN8hPFbkTG2F1AWXGPON5cmm4uiLPfQkWhX/LnClrhzZpNtMJYs5AEFeDs0POijcRugZsQA+wvLi0lSlhOfkqtjAJKpPUwy1wrJFDdvqdQBjpNQh/LB8c15XfQV2JT/3NX26dQe8zvHhL6NvfhBnAikodYkBr7UjSl36CBk0cPebZMZEBBiHdo76xORVkpmqDvkhFByXXeAsvRa2YWS4wxpiNJFswlRhjubGau7LrT113WMcPvgYXHYHf2IYJWD goneri.lebouder@enovance.com"
+        state: present
+        user: zuul
```
pabelanger: same

playbooks/ansible-test-integration-base/run.yaml
```diff
@@ -33,6 +33,11 @@
         _test_options: "{{ _test_options }} --inventory /home/zuul/inventory"
       when: ansible_test_command == 'network-integration'
 
+    - name: Enable --retry-on-error
+      set_fact:
+        _test_options: "{{ _test_options }} --retry-on-error"
+      when: ansible_test_retry_on_error | default(False)
+
```
pabelanger: I don't think we want to do this, if ansible-test cannot successful pass first round, we are likely exposing errors some place.  This will depend on us checking limestone.

This alone, likely means we have to make job non-voting by default.

roles/vmware-ci-nfs-share/tasks/main.yaml
```diff
@@ -0,0 +1,37 @@
...
+- name: Download ISO
+  get_url:
+    url: '{{ item.url }}'
+    dest: '/srv/share/isos/{{ item.name }}'
+    mode: '0644'
+  with_items:
+      - name: fedora.iso
+        url: https://download.fedoraproject.org/pub/fedora/linux/releases/30/Workstation/x86_64/iso/Fedora-Workstation-Live-x86_64-30-1.2.iso
+      - name: centos.iso
+        url: http://isoredirect.centos.org/centos/8/isos/x86_64/CentOS-8-x86_64-1905-dvd1.iso
```
pabelanger: We'll have to cache these on our images going forward, to avoid downloading from web everytime.  This will improve stability.

roles/vmware-ci-set-passwords/files/vcenter_set_password
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/python3
```
pabelanger: Should add LICENSE header

pabelanger: Also, there is no way to do this in ansible today?

goneri: No, not really.

playbooks/ansible-cloud/appliance-base/pre.yaml
goneri: If tThe the appliances are not online, the ssh-keyscan just below will fail.

playbooks/ansible-cloud/vcenter-appliance/files/bootstrap.yaml
goneri: I've just removed them. How can I use the autohold thing?

playbooks/ansible-test-cloud-integration-base/pre.yaml
goneri: Done

playbooks/ansible-test-integration-base/run.yaml
goneri: Yes, yes and no :-)

Until the root cause of the ESXI kernel panic is resolved we will see failure time to time. If the ESXi panic, it will reboot, and so we may have situation were the retry actually succeed.

So yes, it's not elegant, but I would prefer to keep it until the situation is resolved.

roles/vmware-ci-nfs-share/tasks/main.yaml
goneri: Totally agree. Do we already have a solution in place for that?

playbooks/ansible-cloud/appliance-base/pre.yaml
pabelanger: well, the appliance is online but network isn't setup. This is just a safely check, also, if sshd isn't started ssh-keyscan will fail but it will not wait until the port is open.  This logic does that

playbooks/ansible-cloud/vcenter-appliance/files/bootstrap.yaml
```diff
@@ -24,29 +30,17 @@
...
-    # with "vcenter.test==127.0.0.1", this won't work.
-    - name: Write the /etc/hosts file
-      copy:
-        content: |
-          127.0.0.1  localhost
-          ::1  localhost ipv6-localhost ipv6-loopback
-          {{ hostvars['vcenter']['nodepool']['interface_ip']}} vcenter.test vcenter
-        dest: /etc/hosts
-      become: true
+    - name: Change zuul user password (ESXi)
```
pabelanger: understand what we are doing, but we are kinda blurring the line here of this being the 'vcenter bootstrap playbook', can we move this into esxi specific playbooks?

playbooks/ansible-test-cloud-integration-base/pre.yaml
```diff
@@ -9,11 +12,27 @@
...
 
-    - copy:
-        content: |
-            [DEFAULT]
-            vcenter_username: administrator@vsphere.local
-            vcenter_password: !234AaAa56
-            vcenter_hostname: {{ hostvars['vcenter']['nodepool']['interface_ip']}}
-            vmware_validate_certs: false
-        dest: "{{ ansible_user_dir }}/{{ zuul.projects['github.com/ansible/ansible'].src_dir }}/test/integration/cloud-config-vcenter.ini"
+    - import_role:
```
pabelanger: nit, lets add name here to give a little description of what is going on.

playbooks/ansible-test-cloud-integration-base/pre.yaml
```diff
@@ -9,11 +12,27 @@
...
+        name: vmware-ci-nfs-share
+      vars:
+        vmware_ci_nfs_share_allow_ips:
+          - "{{ hostvars['esxi1']['nodepool']['interface_ip'] }}"
+      when: "'esxi1' in hostvars"
+
+    - import_role:
```
pabelanger: same, name info

playbooks/ansible-test-integration-base/run.yaml
pabelanger: I still think this is a blocker, if we are getting panic, the we need to focus on dealing with them. By enabling this, we are just going to ignore IMO

roles/vmware-ci-nfs-share/tasks/main.yaml
```diff
@@ -0,0 +1,38 @@
...
+    path: '{{ item }}'
+    state: directory
+    owner: '{{ vmware_ci_nfs_share_owner_uid }}'
+    group: '{{ vmware_ci_nfs_share_owner_gid }}'
+  with_items:
+    - /srv/share/vms
+    - /srv/share/isos
+  become: true
+
+- name: Download ISO
```
pabelanger: This is a blocker, IMO. Lets add the image to DIB to avoid download cost each job run. I can work on that

zuul.d/ansible-cloud-jobs.yaml
```diff
@@ -40,3 +40,15 @@
...
+
+- job:
+    name: ansible-test-cloud-integration-vcenter_1esxi-python36
+    parent: ansible-test-cloud-integration-vcenter
+    nodeset: vmware-vcsa_1esxi-6.7.0-python36
+    vars:
+      ansible_test_python: 3.6
+      ansible_test_integration_targets: zuul/vmware/vcenter_1esxi/
+      ansible_test_retry_on_error: true
+    attempts: 1
```
pabelanger: this needs to be removed

roles/vmware-ci-nfs-share/tasks/main.yaml
goneri: I use the same roles outside of Zuul. So I would prefer to keep the task. This being said, I'm totally fine with the idea of pre-loading the images.

playbooks/ansible-cloud/vcenter-appliance/files/bootstrap.yaml
goneri: I drop this playbook, it overlaps with `vmware-ci-set-passwords` anyway.

goneri: Regarding the `Set SSH interface IP for appliance` comment, isn't something that nodepool already check?

playbooks/ansible-test-cloud-integration-base/pre.yaml
goneri: Done.

playbooks/ansible-test-cloud-integration-base/pre.yaml
goneri: Done

zuul.d/ansible-cloud-jobs.yaml
goneri: Done

playbooks/ansible-cloud/appliance-base/pre.yaml
goneri: Isn't that something already done by nodepool?

pabelanger: No, this is why in fact we want it. Long term yes we want nodepool to do it for appliances, but not all support it right now. So, keeping this today should be a noop on nodepool, until all appliance are checked via nodepool

goneri: ok, I put it back. Thanks for your patience :-)

playbooks/ansible-cloud/appliance-base/pre.yaml
```diff
@@ -23,7 +23,8 @@
         search_regex: SSH
 
     - name: Ensure remote SSH host keys are known
-      shell: "ssh-keyscan {{ __ssh_ip }} >> ~/.ssh/known_hosts"
+      shell: "ssh-keyscan {{hostvars[item].ansible_host }} >> ~/.ssh/known_hosts"
+      with_inventory_hostnames: appliance
```
pabelanger: why is this needed? We setup __ssh_ip above, which is not used now.

playbooks/ansible-test-cloud-integration-base/pre.yaml
```diff
@@ -9,11 +12,56 @@
...
+
+    - name: Restore the cache files for the NFS datastore
+      block:
+        - name: Create the ISO directory
+          file:
+            path: /srv/share/isos
+            state: directory
+          become: true
+
+        - name: Move the cached files
```
pabelanger: any reason not to symlink or mount /srv/share/ios here?

playbooks/ansible-test-integration-base/run.yaml
pabelanger: Lets drop this, just to see how bad things are.  As I mentioned, I want to iterate on this if an issue. Given job is non-voting, we can see how bad it is.

roles/vmware-ci-configure-ansible-test/templates/cloud-config-vcenter.ini.j2
```diff
@@ -0,0 +1,15 @@
+[DEFAULT]
+vcenter_username: administrator@vsphere.local
+vcenter_password: {{ lookup('file', vmware_ci_set_passwords_secret_dir + '/vcenter/tmp/vcenter_password.txt') }}
```
pabelanger: why does this have to stored on disk and use lookup? This could leak info (which isn't big deal, more curious)

zuul.d/ansible-cloud-jobs.yaml
```diff
@@ -40,3 +40,14 @@
...
+      ansible_test_retry_on_error: true
+
+- job:
+    name: ansible-test-cloud-integration-vcenter_1esxi-python36
+    parent: ansible-test-cloud-integration-vcenter
+    nodeset: vmware-vcsa_1esxi-6.7.0-python36
+    vars:
+      ansible_test_python: 3.6
+      ansible_test_integration_targets: zuul/vmware/vcenter_1esxi/
+      ansible_test_retry_on_error: true
```
pabelanger: can be dropped, when we remove above

roles/vmware-ci-configure-ansible-test/templates/cloud-config-vcenter.ini.j2
goneri: We are dealing with several several Ansible processes, on different hosts. Sharing variables is a giant pain in the neck. I came to the conclusion that this is best approach.

zuul.d/ansible-cloud-jobs.yaml
goneri: You mean, the `retry_on_error`?

playbooks/ansible-cloud/appliance-base/pre.yaml
pabelanger: did see reply here

playbooks/ansible-cloud/vcenter-appliance/files/bootstrap.yaml
```diff
@@ -14,39 +19,7 @@
         key: "{{ _ssh_key }}"
         state: present
         user: zuul
+        path: "{{ authorized_keys_file[inventory_hostname] }}"
```
pabelanger: What is happening here?  Why do we need to change authorized_key file?

zuul.d/ansible-cloud-jobs.yaml
```diff
@@ -40,3 +40,12 @@
     vars:
       ansible_test_python: 3.6
       ansible_test_integration_targets: zuul/vmware/vcenter_only/
+
+- job:
+    name: ansible-test-cloud-integration-vcenter_1esxi-python36
+    parent: ansible-test-cloud-integration-vcenter
```
pabelanger: okay, i think I see the issue here. You've opted to parent to ansible-test-cloud-integration-vcenter over creating an new esxi specific one, which also include vcenter.

The part I need to understand, can esxi be run without vcenter? My gut says, we should likely create ansible-cloud-esxi-appliance. Then, this job parent to both ansible-cloud-vcenter-appliance and ansible-cloud-esxi-appliance.

Stacking it this way, kinda works, but not means that esxi things are in the vcenter playbooks. Which makes review / flow a little messy

goneri: > okay, i think I see the issue here. You've opted to parent to ansible-test-cloud-integration-vcenter over creating an new esxi specific one, which also include vcenter.
> 
> The part I need to understand, can esxi be run without vcenter? My gut says, we should likely create ansible-cloud-esxi-appliance. Then, this job parent to both ansible-cloud-vcenter-appliance and ansible-cloud-esxi-appliance.
> 
> Stacking it this way, kinda works, but not means that esxi things are in the vcenter playbooks. Which makes review / flow a little messy

We can technically start a single ESXi instance and so in this case a `ansible-cloud-esxi-appliance` would make sense. But none of the tests relies on this kind of architecture. We always assume a vcenter is also available. So I took the initiative to just maintain on single playbook. Also, can a job have multiple inheritance?

playbooks/ansible-cloud/vcenter-appliance/files/bootstrap.yaml
goneri: On ESXi, the authorized_keys file is not in `~zuul/.ssh` directory, but in `/etc/ssh`. There is no `/home` directory.

pabelanger:
recheck

pabelanger:
This is likely going to fail the dco-license job now. I just realized, we didn't have that job running for some reason (when we migrated to zuul.a.c). To keep humans happy, you'll need to sign your comments to acknowledged DCO license.

goneri:
Do we really need that? Molecule and  Core don't check the `Signed-off-by` field.

https://github.com/ansible/molecule/issues/1548

pabelanger:
While I am not a lawyer, we either need CLA or DCO to properly license commits.  When this last came up last year, community team agreed the DCO was the way to go. As, that didn't required users to sign legal contracts, and use a light weight --sign-off process in git commits.

https://molecule.readthedocs.io/en/stable/contributing.html#contribution-guidelines even references DCO, even tho it is no longer checked.

AWX also uses DCO, https://www.ansible.com/products/awx-project/faq

What we likely need to do, is contact community again. We can do that first thing monday, but IMO, DCO better then CLA to monitor.

pabelanger:
We cannot control ansible/ansible since we don't run CI there.

goneri:
> While I am not a lawyer, we either need CLA or DCO to properly license commits. When this last came up last year, community team agreed the DCO was the way to go. As, that didn't required users to sign legal contracts, and use a light weight --sign-off process in git commits.
> 
> https://molecule.readthedocs.io/en/stable/contributing.html#contribution-guidelines even references DCO, even tho it is no longer checked.
> 
> AWX also uses DCO, https://www.ansible.com/products/awx-project/faq
> 
> What we likely need to do, is contact community again. We can do that first thing monday, but IMO, DCO better then CLA to monitor.

thanks for the answer. I've just signed the commits.

pabelanger:
I've reverted the job for now, until we make sure everybody knows it is coming. That will allow us to confirm with community too. But yah, hope to enable it shortly

PR #272, opened 2019-12-11
Disable collection testing...

pabelanger:
... until we have _new_ migration jobs working. The current code base
for all collections is about to be refreshed.


pabelanger:
recheck

PR #295, opened 2020-01-09
handle ansible-test with a role

goneri:
Move playbooks/ansible-test-integration-base/run.yaml in a role.


goneri:
See: https://github.com/ansible/ansible-zuul-jobs/pull/360

PR #304, opened 2020-01-13
Fix ansible-tox-molecule

ssbarnea:
- corrects its parent to tox-molecule which also installs docker,
  the most common backend used by molecule
- adds post collection of html report made by tox -e molecule

### DIFF
--- a/zuul.d/jobs.yaml
+++ b/zuul.d/jobs.yaml
@@ -23,8 +23,9 @@
 
 - job:
     name: ansible-tox-molecule
-    parent: tox
+    parent: tox-molecule
     pre-run: playbooks/ansible-tox-molecule/pre.yaml
+    post-run: playbooks/ansible-tox-molecule/post.yaml
     vars:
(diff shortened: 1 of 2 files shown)


pabelanger:
Can you open a PR in the following projects, to confirm nothing breaks (then depends-on this):

https://dashboard.zuul.ansible.com/t/ansible/builds?job_name=ansible-tox-molecule

zuul.d/jobs.yaml
```diff
@@ -23,7 +23,7 @@
 
 - job:
     name: ansible-tox-molecule
-    parent: tox
+    parent: tox-molecule
     pre-run: playbooks/ansible-tox-molecule/pre.yaml
```
pabelanger: will need to add post-run here for playbook above

ssbarnea: Added https://github.com/ansible-security/ids_install/pull/11

PR #311, opened 2020-01-15
Define reusable molecule jobs

ssbarnea:
These jobs should match ansible- ones but also run with docker
enabled.

They rely on YAML 1.1 merge keys feature,... hoping that it will work.
http://blogs.perl.org/users/tinita/2019/05/reusing-data-with-yaml-anchors-aliases-and-merge-keys.html

Needed-By: https://github.com/ansible/molecule/pull/2486

### DIFF
--- a/playbooks/ansible-tox-molecule/pre.yaml
+++ b/playbooks/ansible-tox-molecule/pre.yaml
@@ -1,5 +1,20 @@
 ---
 - hosts: all
   tasks:
+
     - name: Reset SSH connection for new group
       meta: reset_connection
+
+    # TODO(ssbarnea): remove hack once upstream patch merges:
+    # https://review.opendev.org/#/c/703053/
+    - name: Workaround for yum --nobest with docker-ce
+      when:
+        - ansible_distribution == "CentOS"
+        - ansible_distribution_major_version == "8"
+      become: true
+      yum:
+        name: https://download.docker.com/linux/centos/7/x86_64/stable/Packages/containerd.io-1.2.6-3.3.el7.x86_64.rpm
+
+    - name: Install docker
+      include_role:
+        name: install-docker
--- a/zuul.d/jobs.yaml
+++ b/zuul.d/jobs.yaml
@@ -22,10 +22,26 @@
     nodeset: centos-8-1vcpu
 
 - job:
-    name: ansible-tox-molecule
-    parent: tox-molecule
+    name: ansible-tox-docs
+    parent: tox-docs
+    nodeset: centos-8-1vcpu
+
+- job: &ansible-tox-molecule-base
+    name: ansible-tox-molecule-base
+    # Stub job used to define only properties specific to molecule jobs,
+    # expected to decorate other job definitions via anchors. To avoid overrides
+    # it does *not* have: nodeset, abstract, description
     pre-run: playbooks/ansible-tox-molecule/pre.yaml
     post-run: playbooks/ansible-tox-molecule/post.yaml
+    timeout: 5400  # 1.5h as they are expected to be slower by nature
+
+- job:
+    <<: *ansible-tox-molecule-base
+    name: ansible-tox-molecule
+    description: |
+      Generic molecule jobs that does not specify which python version will
+      be using, recommended when that is not important.
+    parent: tox-molecule
     vars:
       tox_envlist: molecule
     nodeset: fedora-latest-1vcpu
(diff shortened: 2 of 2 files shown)


pabelanger:
recheck

pabelanger:
This was due to centos8.1 release, as a result we've rebuilt our images, and should now have centos8.1 nodes online

ssbarnea:
recheck

zuul.d/jobs.yaml
```diff
@@ -588,3 +589,49 @@
     nodeset: vyos-1.1.8-python38
     vars:
       ansible_test_python: 3.8
+
+# molecule- prefixed jobs match ansible- ones but also enable extra
+# virtualization services that are use by molecule, like docker, podman (TBD)
+# or libvirt (TBD).
+
+- job:
+    <<: [*ansible-tox-molecule, *ansible-tox-linters]
```
goneri: Hi,

Why don't you use the `parent` key instead to inheritance from the other job? I feel like this kind of advanced YAML incantation don't really improve the readability.

ssbarnea: Zuul does not supports multiple inheritance and in that case we need to combine properties from 3 sources: 
- python versions/nodesets (each tox pyXY job has a specific nodeset)
- main tox-molecule job which adds the pre-run that installs docker/podman
- custom tox environmnets

As you can see is not merged yet, but apparently merging anchors worked as desired.

The main benefit here is that a setup like this allow use to contro the nodeset from a single place, or how the molecule pre-run yaml from a single location too.

@goneri You are welcomed to experiment with others setups maybe we find a better setup. I agree that these are not very easy to read but they achieve the goal of avoiding configs from diverging, something that happened a log on with triple jobs. I even have small script that dumps the expanded yaml.

ssbarnea: Indeed lack of visible (implicit) parent is a source of concern here. That is WIP, so I will continue to play with it and until i bring it in an easy to maintain shape.

If i remember well there was a problem with overriding order but now that I know that this setup works, I can save it and experiment with other setups and diff the results, expanded YAML should remain the same. Maybe I can scrap the double dictionary merge.

playbooks/ansible-tox-molecule/pre.yaml
```diff
@@ -1,5 +1,18 @@
...
     - name: Reset SSH connection for new group
       meta: reset_connection
+
+    - name: Workaround for yum --nobest with docker-ce
+      when:
+        - ansible_distribution == "CentOS"
+        - ansible_distribution_major_version == "8"
+      become: true
+      yum:
+        name: https://download.docker.com/linux/centos/7/x86_64/stable/Packages/containerd.io-1.2.6-3.3.el7.x86_64.rpm
```
pabelanger: I'd like to see us link to some sort of upstream issue around this, as we likely should fix this in the install-docker role.

zuul.d/jobs.yaml
```diff
@@ -22,10 +22,26 @@
     nodeset: centos-8-1vcpu
 
 - job:
-    name: ansible-tox-molecule
-    parent: tox-molecule
+    name: ansible-tox-docs
```
pabelanger: Lets remove this here, I believe I have another PR up for this.

zuul.d/jobs.yaml
```diff
@@ -22,10 +22,26 @@
...
+    name: _stub_molecule
+    # Stub job used to define only properties specific to molecule jobs,
+    # expected to decorate other job definitions via anchors. To avoid overrides
+    # it does not have: nodeset, abstract, description
     pre-run: playbooks/ansible-tox-molecule/pre.yaml
     post-run: playbooks/ansible-tox-molecule/post.yaml
+    timeout: 5400  # 1.5h as they are expected to be slower by nature
+
+- job:
+    <<: *_stub_molecule
```
pabelanger: I think molecule-tox-base would be better, and fix into current naming scheme.

zuul.d/jobs.yaml
```diff
@@ -22,10 +22,26 @@
...
+    name: ansible-tox-molecule
+    description: |
+      Generic molecule jobs that does not specify which python version will
+      be using, recommended when that is not important.
```
pabelanger: You maybe should set job.abstract = true here, so people don't use it directly

ssbarnea: There is nothing wrong in using this job, in fact is recommented for most users. The most specific ones are needed for special cases, like testing molecule itself. But when someone wants to just run molecule, they usually do not care which python version is used. I would have added abstract if I wanted to avoid its use.

zuul.d/jobs.yaml
ssbarnea: The reson why I used `_stub_molecule` is to follow python private var pattern (underline prefix) because I am unable to use abstract on it.
Also I wanted to avoid poluting list of job in job-tree in zuul. Thus is why opeted for that name, making it more obvious that they should never try to use that partial/incomplet job.

Whatever name we pick, it should better start with a prefix that indicates that is not a normal job. I wonder if we can use dot, like on linux. Bad part about dot is that is listed before "a".

If we are to look read, https://stackoverflow.com/questions/8086375/what-character-to-use-to-put-an-item-at-the-end-of-an-alphabetic-list we are better of using `z_`, I think.

playbooks/ansible-tox-molecule/pre.yaml
ssbarnea: follow up on https://review.opendev.org/#/c/703053/

zuul.d/jobs.yaml
```diff
@@ -22,10 +22,26 @@
     nodeset: centos-8-1vcpu
 
 - job:
-    name: ansible-tox-molecule
-    parent: tox-molecule
+    name: ansible-tox-docs
+    parent: tox-docs
+    nodeset: centos-8-1vcpu
+
+- job: &_stub_molecule
```
pabelanger: moving to here, as this is the actual job.

I don't think I want to get into the idea of private / public jobs, while I understand the idea. Until now, with parent / child jobs' we've used the idea of 'base' to indicate a top-level job, where child to parent to.

This is a little different, as we are trying reduce copypasta in job design. While not fully documented, zuul jobs can have multiple parents, so I wonder if that might be something to try here.  Let me experiment with something in etherpad, so see if we can review all options.

pabelanger: This would be the zuul way to do this, without using yaml anchors, which means a little more typing, but removes need for private jobs:

  https://etherpad.openstack.org/p/ansible-tox-molecule

PR #334, opened 2020-01-27
configure-swap: create the swapfile only once

goneri:
Do not try (and fail) to recreate the swapfile if it already exists.

Signed-off-by: Gonéri Le Bouder <goneri@lebouder.net>

### DIFF
--- a/roles/configure-swap/tasks/main.yaml
+++ b/roles/configure-swap/tasks/main.yaml
@@ -1,7 +1,13 @@
 ---
+- name: Check if swapfile exists
+  stat:
+    path: /root/swapfile
+  become: true
+  register: swapfile_result
 - include: root.yaml
   static: false
   when:
+    - not swapfile_result.stat.exists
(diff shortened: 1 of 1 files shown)


roles/configure-swap/tasks/main.yaml
```diff
@@ -1,7 +1,13 @@
 ---
+- name: Check if swapfile exists
+  stat:
+    path: /root/swapfile
+  become: true
+  register: swapfile_result
 - include: root.yaml
   static: false
   when:
+    - swapfile_result.stat.exists == False
```
pabelanger: not swapfile_result.stat.exists

roles/configure-swap/tasks/main.yaml
```diff
@@ -1,7 +1,13 @@
 ---
+- name: Check if swapfile exists
+  stat:
+    path: /root/swapfile
+  become: true
+  register: swapfile_result
```
pabelanger: do you mind adding space between tasks? I find is much easier to review

PR #338, opened 2020-01-29
vcenter_1esxi-python36: bump timeout to 10800s

goneri:
The last run with the previous timeout has failed during
`vmware_vm_info` test.


zuul.d/ansible-cloud-jobs.yaml
```diff
@@ -48,7 +48,8 @@
     vars:
       ansible_test_python: 3.6
       ansible_test_integration_targets: zuul/vmware/vcenter_1esxi/
-    timeout: 10000
+    # 5h
```
pabelanger: This isn't 5h, but 3h

PR #351, opened 2020-02-10
vmware: Write the configuration for collection testing

goneri:


playbooks/ansible-test-integration-base/run.yaml
```diff
@@ -91,8 +91,10 @@
           set_fact:
             _test_location: "~/.ansible/collection/ansible_collections/{{ _collection_namespace.stdout }}/{{ _collection_name.stdout }}"
 
-        - name: Copy ansible.cfg to correct location
-          shell: "cp {{ ansible_user_dir }}/{{ zuul.projects['github.com/ansible/ansible'].src_dir }}/test/integration/{{ ansible_test_command }}.cfg {{ _test_location }}/tests/integration/{{ ansible_test_command }}.cfg"
```
pabelanger: I'd prefer if we didn't change this, but we add a take that follows that only applied to 'integration' jobs. See https://github.com/ansible/ansible-zuul-jobs/blob/master/playbooks/ansible-test-integration-base/run.yaml#L34 for example

PR #360, opened 2020-02-26
handle ansible-test with a role

goneri:
Move playbooks/ansible-test-integration-base/run.yaml in two new roles:

- deploy-artifacts: fetch and install the artifacts (sdist and
  collections)
- ansible-test: run ansible-test

Signed-off-by: Gonéri Le Bouder <goneri@lebouder.net>


playbooks/ansible-test-integration-base/run.yaml
```diff
@@ -1,114 +1,14 @@
...
-    - name: Run the integration test suite
-      args:
-        chdir: "{{ _test_location }}"
-        executable: /bin/bash
-      environment: "{{ ansible_test_environment | default({}) }}"
-      shell: "source ~/venv/bin/activate; {{ _test_executable }} {{ ansible_test_command }} {{ _test_options }} --python {{ ansible_test_python }} -vvvv {{ ansible_test_integration_targets }}"
+    - name: Fetch and install the artifacts
+      import_role:
+        name: deploy-artifacts
+    - name: Run ansible-test
```
pabelanger: space here for readability please.

roles/ansible-test/tasks/main.yaml
```diff
@@ -0,0 +1,33 @@
...
+  import_tasks: init_test_options.yaml
+
+- name: Check if galaxy.yml exists
+  stat:
+    path: "{{ ansible_test_location }}/galaxy.yml"
+  register: galaxy_stat_result
+
+- name: Single ansible mode
+  import_tasks: init_collection.yaml
+  when: not galaxy_stat_result.stat.exists
```
pabelanger: lets keep ansible_test_collections for this check, not the galaxy file. As there is going to be settings we need to toggle in ansible-test options and don't want to always check if galaxy.yaml exists.

roles/ansible-test/tasks/main.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+- name: Ansible with collections
+  when: galaxy_stat_result.stat.exists
```
pabelanger: same: ansible_test_collections

roles/ansible-test/tasks/main.yaml
```diff
@@ -0,0 +1,33 @@
...
+- name: Include branch specific variables.
+  include_vars: "{{ item }}"
+  with_first_found:
+    - '{{ ansible_test_git_branch }}.yaml'
+    - default.yaml
+
+- name: Prepare ansible-test parameters
+  import_tasks: init_test_options.yaml
+
+- name: Check if galaxy.yml exists
```
pabelanger: we can remove this and use ansible_test_collections

roles/ansible-test/tasks/main.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+- name: Single ansible mode
+  import_tasks: init_collection.yaml
+  when: not galaxy_stat_result.stat.exists
+
+- name: Ansible with collections
+  when: galaxy_stat_result.stat.exists
+  import_tasks: init_collection.yaml
+
+- name: Install ara
```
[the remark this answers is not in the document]
goneri: This is not really better than the current situation, e.g: the ansible 2.9 -> ansible-minimal 2.10 issue. So yes, I agree, the venv approach is probably the best one.

roles/ansible-test/tasks/main.yaml
goneri: Ok, I don't mind putting that back. It tried to reduce the number of role parameter. I'm under the impression we will only test collection soon. @pabelanger, Am I wrong?

roles/ansible-test/tasks/main.yaml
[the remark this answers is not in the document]
pabelanger: lets revert this change, for now. We can then get the ara-env working, then move into role

roles/ansible-test/tasks/main.yaml
pabelanger: Yes, soon. We still need to test stable-2.9 and stable-2.8, but we can flip the logic once ansible/ansible devel has been unfrozen

goneri: Oh. I forgot about the old stable branches.

playbooks/ansible-test-integration-base/pre.yaml
```diff
@@ -29,24 +29,3 @@
...
-        section: persistent_connection
-        option: log_messages
-        value: true
-
-    - name: Set ansible log path
-      ini_file:
-        path: "{{ ansible_user_dir }}/{{ zuul.projects['github.com/ansible/ansible'].src_dir }}/test/integration/{{ ansible_test_command }}.cfg"
-        section: defaults
-        option: log_path
-        value: ~/ansible-debug.txt
```
pabelanger: lets skip this change for now, I don't think that breaks anything in the role ATM.

roles/ansible-test/tasks/main.yaml
```diff
@@ -0,0 +1,28 @@
...
+  import_tasks: init_collection.yaml
+  when: not ansible_test_collections
+
+- name: Ansible with collections
+  import_tasks: init_collection.yaml
+  when: ansible_test_collections
+
+- name: Install ara
+  import_tasks: install_ara.yaml
+  when: ansible_test_enable_ara
```
pabelanger: This can be remove too, if we do above

playbooks/ansible-test-integration-base/run.yaml
```diff
@@ -1,114 +1,15 @@
...
+      import_role:
+        name: ansible-test
+      vars:
+        ansible_test_test_command: "{{ ansible_test_command }}"
+        ansible_test_location: "{{ ansible_user_dir }}/{{ zuul.project.src_dir }}"
+        ansible_test_git_branch: "{{ zuul.projects['github.com/ansible/ansible'].checkout }}"
+        ansible_test_enable_ara: true
```
pabelanger: same

roles/ansible-test/tasks/main.yaml
goneri: We still need to configure Ansible.

playbooks/ansible-test-integration-base/run.yaml
goneri: We still need to enable Ara.

playbooks/ansible-test-integration-base/pre.yaml
goneri: Why?

pabelanger: I was going to push up the PR to move ARA into its own venv. Which is going on conflict on this. 

My concern as now, is we change the dependency order a little, also minimized the change

goneri: I will be happy to do the Ara role, but I also want to see this one merged...

PR #373, opened 2020-03-16
import the Ara role

goneri:
Isolate all the Ara operations in a new role.


goneri:
This won't fly because Ansible load the callback. If there is no `ara` lib in the python path, the loading with just fail. Ara has to be installed in the same venv.

PR #375, opened 2020-03-19
Make ansible-collections-cisco-asa Python35 jobs Non-Voting

justjais:
Make ansible-collections-cisco-asa Python35 jobs Non-Voting as it's getting to Retry_Limit state every-time, which is happening coz Cisco ASA comes with `diffie-hellman-group14-sha256` and that isn't supported on Ubuntu Xenial where currently python35 setup is configured. So making py35 non-voting for the PR to be allowed to merged.

### DIFF
--- a/zuul.d/project-templates.yaml
+++ b/zuul.d/project-templates.yaml
@@ -86,7 +86,7 @@
               ansible_test_collections: true
         - ansible-test-network-integration-asa-python35:
             vars:
-              ansible_test_collections: true
+              ansible_test_collections: false
         - ansible-test-network-integration-asa-python36:
             vars:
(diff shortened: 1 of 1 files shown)


zuul.d/project-templates.yaml
```diff
@@ -86,7 +86,7 @@
               ansible_test_collections: true
         - ansible-test-network-integration-asa-python35:
             vars:
-              ansible_test_collections: true
+              ansible_test_collections: false
```
pabelanger: Don't change this, add voting: false like:

  https://github.com/ansible/ansible-zuul-jobs/blob/master/zuul.d/project-templates.yaml#L180

PR #390, opened 2020-03-30
deploy-artifacts: ensure runme.sh scripts are executable

goneri:
The +x flag of the files are lost in the tarballs generated by
`ansible-galaxy`. We need to put them back, or the associated
tests will fail.


goneri:
I will revert the patch as soon as the backport is released (https://github.com/ansible/ansible/pull/68451).

PR #391, opened 2020-03-31
Add podman to ansible-tox-molecule job

ssbarnea:
Needed-By: https://github.com/ansible-community/molecule/pull/2632

### DIFF
--- a/playbooks/ansible-tox-molecule/pre.yaml
+++ b/playbooks/ansible-tox-molecule/pre.yaml
@@ -18,3 +18,44 @@
     - name: Install docker
       include_role:
         name: install-docker
+
+    - name: Validate docker engine
+      # on purpose to verify that non-root user can call docker/podman
+      become: false
+      shell: |
+        set -eux
+        newgrp docker || true
+        docker version
+        docker info
+        docker ps
+
+    - name: Install podman (Debian)
+      when:
+        - ansible_os_family != "RedHat"
+      include_role:
+        name: install-podman
+
+    # TODO(ssbarnea): Migrate logic to install-podman (zuul-roles)
+    - name: Install podman (RedHat)
+      # Podman on centos-7 cannot run as non-root and kernel.unprivileged_userns_clone is missing from its kernel
+      # https://github.com/containers/libpod/issues/3207
+      when:
+        - ansible_os_family == "RedHat"
+        - not (ansible_distribution == "CentOS" and ansible_distribution_major_version == "7")
+      become: true
+      package:
+        name:
+          - podman
+          - conmon
+
+    - name: Validate podman engine
+      # on purpose to verify that non-root user can call docker/podman
+      when:
+        - not (ansible_distribution == "CentOS" and ansible_distribution_major_version == "7")
+      become: false
+      shell: |
+        set -eux
+        podman version
+        podman info
+        podman ps
+      changed_when: false


ssbarnea:
@goneri @pabelanger please gate this, the centos-7 is known broken unrelated to this change.

zuul.d/project.yaml
```diff
@@ -44,6 +44,18 @@
...
+              - ^bindep.txt$
+              - ^requirements.txt$
+              - ^test-requirements.txt$
+              - ^tox.ini$
+              - playbooks/ansible-tox-molecule/.*
+            vars:
+              tox_extra_args: --notest
+        - molecule-tox-py36-centos-8: *tox
+        - molecule-tox-py36-ubuntu-bionic: *tox
+        - molecule-tox-py37-fedora-30: *tox
```
pabelanger, line 58: I am unsure we want to add these jobs here. I wonder if we should maybe consider shared zuul jobs for molecule? Which then you and molecule test directly control

ssbarnea: In fact these jobs do not have any "molecule" on them but they add "docker" and "podman" which is needed by molecule jobs.

These jobs are already used in other projects, but they are defined in molecule one. I have nothing against moving them under ansible-zuul-jobs if you want, as I see them as multi-purpose.

Still there is one thing that I will continue to want: to run them when we update the pre.yaml file, so we can easily confirm that the patch is correct. This is why I had the special pattern added, as I did not want to trigger them on random changes, only on those that could break them.

playbooks/ansible-tox-molecule/pre.yaml
```diff
@@ -18,3 +18,44 @@
     - name: Install docker
       include_role:
         name: install-docker
+
+    - name: Validate docker engine
```
goneri, line 22: I would put that in the `install-docker` role.

playbooks/ansible-tox-molecule/pre.yaml
```diff
@@ -18,3 +18,44 @@
...
+        name: install-podman
+
+    # TODO(ssbarnea): Migrate logic to install-podman (zuul-roles)
+    - name: Install podman (RedHat)
+      # Podman on centos-7 cannot run as non-root and kernel.unprivileged_userns_clone is missing from its kernel
+      # https://github.com/containers/libpod/issues/3207
+      when:
+        - ansible_os_family == "RedHat"
+        - not (ansible_distribution == "CentOS" and ansible_distribution_major_version == "7")
+      become: true
```
goneri, line 45: Why isn't this in `install-podman` role?

playbooks/ansible-tox-molecule/pre.yaml
```diff
@@ -18,3 +18,44 @@
...
+      package:
+        name:
+          - podman
+          - conmon
+
+    - name: Validate podman engine
+      # on purpose to verify that non-root user can call docker/podman
+      when:
```
goneri, line 53: I would also put that in the `install-podman`.

ssbarnea:
The answer to most question is because merging changes to "zuul-roles" takes time, sometimes a lot of time. Take a look at https://review.opendev.org/#/c/716578/

I am in favour of making all changes upstream, and in many cases I start by proposing them there before even making them here. Still, gettings things "done" is a different thing and the lack of podman on this instance become a blocker for molecule.

goneri:
> The answer to most question is because merging changes to "zuul-roles" takes time, sometimes a lot of time. Take a look at https://review.opendev.org/#/c/716578/
> 
> I am in favour of making all changes upstream, and in many cases I start by proposing them there before even making them here. Still, gettings things "done" is a different thing and the lack of podman on this instance become a blocker for molecule.

Yeah, I agree. Personally, I'm not sure it makes sense to waste^wspend time upstream, for this kind of thing. I would just prepare a new role in ansible-zuul-jobs/roles and consume it.

ssbarnea:
@goneri @pabelanger I removed any job testing from the change to address your concerns regarding breakage risks. There is refactoring work pending on zuul-roles that affects both install-docker and install-podman and as soon the changes land there I will propose another change to remove our own code from here.

Please help me merge this because it already blocked molecule patches for more than a week and I want to gate several of them over the weekend.

PR #401, opened 2020-04-03
vmware: govcsim integration

goneri:
Depends-On: https://github.com/ansible-collections/vmware/pull/95
    
- starts govcsim on 3 different nodes
- wait for govcsim success before starting the other jobs, on regular lab


goneri:
recheck

PR #412, opened 2020-04-09
Fix vmware gosim job

pabelanger:


goneri:
Thanks @pabelanger! What was the problem?

PR #417, opened 2020-04-16
Install ARA into it's own venv

pabelanger:
This allows use to remove the ARA dependencies from ansible-test
environment.


goneri:
I've closed https://github.com/ansible/ansible-zuul-jobs/pull/373 because this was not working as expected. When Ansible tries to load the Ara callback, it needs to load the `ara` module. And the loading fails because the `ara` module is only available... in the Ara venv.

PR #424, opened 2020-04-21
Enable unit tests on CI for arista.eos 

GomathiselviS:
Unit testcases of arist.eos will run as part of CI

### DIFF
--- a/zuul.d/jobs.yaml
+++ b/zuul.d/jobs.yaml
@@ -298,6 +298,16 @@
       ansible_test_integration_targets: "eos_.*"
 
 - job:
+    name: ansible-test-units-eos
+    parent: ansible-test-units-base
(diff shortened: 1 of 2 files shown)


zuul.d/jobs.yaml
```diff
@@ -297,6 +297,16 @@
       ansible_test_command: network-integration
       ansible_test_integration_targets: "eos_.*"
 
+- job:
+    name: ansible-test-units-eos
+    parent: ansible-test-units-base
+    required-projects:
+      - name: github.com/ansible-collections/arista.eos
+    timeout: 3600
```
pabelanger, line 305: we skip this, as we set it in the parent. We can remove it in the next PR, same for openvswitch

PR #437, opened 2020-04-27
Split NX-OS jobs

NilashishC:

### DIFF
--- a/zuul.d/project-templates.yaml
+++ b/zuul.d/project-templates.yaml
@@ -187,7 +187,19 @@
     name: ansible-collections-cisco-nxos
     check:
       jobs:
-        - ansible-test-network-integration-nxos-cli-python36:
+        - ansible-test-network-integration-nxos-cli-python36_1_of_4:
+            vars:
+              ansible_test_collections: true
+            voting: false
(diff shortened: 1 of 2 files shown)


zuul.d/project-templates.yaml
```diff
@@ -187,10 +187,21 @@
...
             voting: false
+        - ansible-test-network-integration-nxos-cli-python36_2_of_4:
+            vars:
+              ansible_test_collections: true
+            voting: false
+        - ansible-test-network-integration-nxos-cli-python36_3_of_4:
+            vars:
+              ansible_test_collections: true
+            voting: false
+        - ansible-test-network-integration-nxos-cli-python36_4_of_4:
```
pabelanger, line 202:
```
voting: false
```
We need to set voting false here

zuul.d/project-templates.yaml
```diff
@@ -187,10 +187,22 @@
...
+            vars:
+              ansible_test_collections: true
+              voting: false
```
pabelanger, line 205: this is wrong, too many tabs. see above

NilashishC: fixed

PR #443, opened 2020-05-01
ansible-lint: fix docs location

ssbarnea:
Ansible-lint does not produce sphinx docs in the standard location expected by Zuul, so we help
him find them.

Needed-By: https://github.com/ansible/ansible-lint/pull/754


ssbarnea:
recheck

ssbarnea:
recheck

ssbarnea:
Please workflow ASAP (add gate label), ignoring the 3rd party is ok, it was not even supposed to run on this change. These jobs are used only by ansible-lint repository and we got confirmation that the change fixes the problem.

PR #454, opened 2020-05-08
Increase junos vsrx job timeout

danielmellado:
This commit bumps the job timeout for vsrx to 3h.


pabelanger:
no longer needed, we've refactored jobs

PR #456, opened 2020-05-11
unbound: ensure the package is installed

goneri:
In the case of AWS, we reuse existing generic images. They are not
generated by DIB (yet!). So, some assumption regarding the state
of the image may be wrong.


pabelanger:
Can you add depends on here to another openvswitch job, lets say.  Long term, I would like to revert this change, but if working it unblocks aws

goneri:
Yes, I agree.

I created PR on openvswitch and the CI pass: https://github.com/ansible-collections/openvswitch.openvswitch/pull/8

So I merge the patch.

PR #457, opened 2020-05-11
bindep: ensure virtualenv is installed

goneri:
Currently, we use the generic AWS images and some dependencies are
missing when we compare with those generated by DIB.


goneri:
Supersede by https://github.com/ansible/ansible-zuul-jobs/pull/458

PR #458, opened 2020-05-11
ensure virtualenv is installed before bindep run

goneri:
Bindep uses the `virtualenv` command.

### DIFF
--- a/playbooks/ansible-network-appliance-base/pre.yaml
+++ b/playbooks/ansible-network-appliance-base/pre.yaml
@@ -1,6 +1,13 @@
 ---
 - hosts: controller
   tasks:
+    - name: Ensure virtualenv package is installed -- See https://review.opendev.org/#/c/727413/
+      package:
+        name: virtualenv
+        state: present
+      become: true
+      when: "'aws' in group_names"
+
     - name: Install binary dependencies
       include_role:
         name: bindep


playbooks/ansible-network-appliance-base/pre.yaml
```diff
@@ -1,6 +1,12 @@
 ---
 - hosts: controller
   tasks:
+    - name: Ensure virtualenv package is installed -- See https://review.opendev.org/#/c/727413/
```
pabelanger, line 4: So for this, I think we it would be best to make this condition on AWS.  One idea I have, is we create a nodeset group_var, for appliance but some how reference aws.  That way, this only runs on aws nodes, and not modify DIBs.

goneri: What do we need to make the AWS node reach parity with a DIB image?

pabelanger: realistically, take DIB image, export as AMI and upload it. That would be best solution, but that required work to happen.

That's why I think making this condition is a good step to avoid unneeded DIB config changes

goneri: Yeah, I agree. Now the block is only enabled on `aws`, so we will avoid any side effects.

PR #451, opened 2020-05-07
prepare the SplunkEnterprise-SES-8.0.0-python38 job

goneri:
Depends-On: https://github.com/ansible/ansible-zuul-jobs/pull/471

Based on `ansible-network-appliance-base`.

PR #459, opened 2020-05-12
bionic ensure py3{6,8}-dev is installed

goneri:
AWS images don't use DIB, and so we need to pull these dependencies.


bindep.txt
```diff
@@ -8,7 +8,11 @@ libssh-dev [test platform:dpkg]
 libxml2-dev [test platform:ubuntu-bionic]
 libxslt1-dev [test platform:ubuntu-bionic]
 python2.7-dev [test platform:dpkg]
+python3.6-dev [test platform:ubuntu-bionic]
```
pabelanger, line 11: What failure does this fix?  I assume something doesn't compile properly because python is missing.  Which version do we need on bionic?

goneri: For now, I only create jobs for 3.6 and 3.8:

https://github.com/ansible/ansible-zuul-jobs/pull/451/files#diff-e0a9ae430091fa99c8f126c4410bc32dR32

pabelanger: k, lets keep 3.6 for bionic, 3.7 can be dropped. We already installed 3.8 via ansible, based on ansible-test var.

goneri: Roger that.

PR #460, opened 2020-05-12
bindep: ensure tox is installed

goneri:
The AWS images are not build using DIB yet. Instead we use the generical
images. Those don't come with `tox` by default, we need to pull the
package.


goneri:
Note: I don't think this should be a requirement, because of https://opendev.org/zuul/zuul-jobs/src/branch/master/playbooks/tox/pre.yaml#L4

goneri:
> Note: I don't think this should be a requirement, because of https://opendev.org/zuul/zuul-jobs/src/branch/master/playbooks/tox/pre.yaml#L4

Answering to myself. This link point on a playbook, we actually call the role directly, so these tasks never get run.

goneri:
Supersede by: https://github.com/ansible/ansible-zuul-jobs/pull/463

PR #463, opened 2020-05-13
call ensure-tox role before tox

goneri:
The `tox` role assumes the command is available.


playbooks/ansible-test-network-integration-base/pre.yaml
```diff
@@ -11,6 +11,10 @@
         _network_os: "{{ hostvars[groups['openvswitch'][0]]['ansible_network_os'] }}"
       when: '"openvswitch" in groups'
 
+    - name: Ensure tox
```
pabelanger, line 14: I feel this is fine, again we should only need to limit to aws nodes. As tox is managed via DIB.  The fear I have, is we overwrite DIBs setup and manage if tox

goneri:
We should not need that anyway thanks to https://review.opendev.org/#/c/727561/

PR #490, opened 2020-05-29
Set junos unit test job to non-voting

danielmellado:
Before the dependencies on the job [1] are merged, triggering unit test
will just fail. It seems that until recently ansible-test was providing
such dependencies, so we were able to merge jobs, but that's no longer
the case.

Once [1] is merged, we'll reenable the voting

[1] https://github.com/ansible-collections/junipernetworks.junos/pull/52


pabelanger:
This is no longer needed

PR #499, opened 2020-06-03
ci: ansible-cloud-vcenter-appliance voting

goneri:
- Trigger `ansible-cloud-vcenter-appliance` only when needed.
- Make it voting.


zuul.d/project.yaml
```diff
@@ -5,7 +5,10 @@
     check:
       jobs:
         - ansible-cloud-vcenter-appliance:
-            voting: false
+            files:
```
pabelanger, line 8: also need in gate pipeline

PR #501, opened 2020-06-04
nodesets: add appliance_no_ssh group

goneri:
`add-build-sshkey` from `base-minimal` will regenerate a new set
of SSH key and push them on the nodes. However it ignores the
hosts from the `appliance` group. As a result, there is no way to
access the appliances over SSH after this step.

The `appliance` group is too broad. This new group resolves the
problem. It clearly identifies the hosts that should not be targeted
by this operation.

We currently face the problem with the vcenter and splunk hosts. So
far the workaround was to hardcode a known password to the zuul users.

### DIFF
--- a/zuul.d/nodesets.yaml
+++ b/zuul.d/nodesets.yaml
@@ -81,6 +81,9 @@
         nodes:
           - vcenter
           - esxi1
+      - name: appliance_no_ssh
+        nodes:
+          - esxi1
       - name: esxis
         nodes:
           - esxi1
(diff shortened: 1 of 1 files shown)


pabelanger:
So, I think you are right, but I actually think we should invert this. By default, anything in the 'appliance' group, won't have ssh.  But we _should_ add a new 'appliance-ssh' group, and move nodesets that support it into it.  For example, the openvswitch nodeset too.

This way, we can split https://github.com/ansible/project-config/blob/master/playbooks/base-minimal-test/pre.yaml#L15 into something like:

hosts: all:appliance-ssh:!appliance

And replace any reference to openvswitch group with appliance-ssh

goneri:
> hosts: all:appliance-ssh:!appliance

This won't work IMO, `all` already include `appliance-ssh`, and `appliance` include `appliance-ssh`. This is why I substract the non-ssh hosts from `all`.

PR #503, opened 2020-06-04
vmware: use appliance-ssh group

goneri:
Depends-On: https://github.com/ansible/ansible-zuul-jobs/pull/507

Use the `appliance-ssh` for the vcenter host. It's a regular Linux VM
and we can update the authorized_keys with `prepare-workspace-git`.


goneri:
recheck

goneri:
recheck

PR #504, opened 2020-06-04
openvswitch: use appliance-ssh to target appliance

goneri:
Depends-On: https://github.com/ansible/ansible-zuul-jobs/pull/507

We apply `prepare-workspace-git` too all host, but the ones from
the `appliance` groups. As a result, all the hosts from the `appliance`
won't have the temporary SSH key, and won't be reachable over SSH.

Until now we were using an `openvswitch` to target the openvswitch
appliance. This way, they were not part of the `appliance` group.
And so, the SSH key was update.

We now have an `appliance-ssh` group for this purpose. This commit
makes us of it.


goneri:
recheck

goneri:
recheck

PR #515, opened 2020-06-08
ansible-test: run sanity test git repository

goneri:
So... `ansible-test sanity` has to be run in a `git` repository. We
assume `ansible_test_location` is a git repository and run the command
there.

### DIFF
--- a/roles/ansible-test/tasks/init_collection.yaml
+++ b/roles/ansible-test/tasks/init_collection.yaml
@@ -19,6 +19,12 @@
     _test_location: "{{ ansible_test_collection_dir }}/{{ _collection_namespace.stdout }}/{{ _collection_name.stdout }}"
     _test_cfg_location: "{{ ansible_test_collection_dir }}/{{ _collection_namespace.stdout }}/{{ _collection_name.stdout }}/tests/integration/{{ ansible_test_test_command }}.cfg"
 
+- name: sanity test should be run from the git repository - https://github.com/ansible/ansible/issues/69953
+  shell: |
+    rm -r {{ _test_location }}
+    cp -r {{ ansible_test_location }} {{ _test_location }}
+  when: ansible_test_test_command == 'sanity'
+
 - name: Install python requirements
   shell: "{{ ansible_test_venv_path }}/bin/pip install -r {{ _test_location }}/requirements.txt -r {{ _test_location }}/test-requirements.txt"
 


pabelanger:
This is not correct, by default ansible-test sanity will not work on a git repo, it should be the installed collection.

You can see this working on https://c840b10bdb1c4ffd5f46-d4b4e639432347957be7d1e5f737e888.ssl.cf2.rackcdn.com/3/0a98184c5fcd07009cb520e5f8df76327169d19c/check/ansible-test-sanity-splunk/b39aa5a/job-output.html

Which issue are you trying to fix?

goneri:
The collection should be a git repository, if we want things to work fine. I use this hack to "fix" that:
https://github.com/ansible/ansible/issues/69953

- without the hack, `validate-modules` raises several errors:
https://dashboard.zuul.ansible.com/t/ansible/build/b39aa5a908ea4733825fa79cc40cd45b
- with the hack, there is still a pylint error but the other is gone: https://dashboard.zuul.ansible.com/t/ansible/build/ad498201ad6641238882d3b5bf919dd1

pabelanger:
I thought more about this today, I feel we don't want to do this. If we upload the collection to automation hub, and it runs ansible-test sanity using the installed collection, this is still going to fail.

This morning, we should confirm automation hub will run sanity from the collection, if it will, then we don't want to do this.

pabelanger:
yes, AH is using the collection tarball, not git repos. So, I feel we cannot do this, otherwise tests will pass in zuul, but have the same failure in AH.

https://github.com/ansible/galaxy-importer/blob/master/docker/ansible-test/entrypoint.sh

We should fix collections to work properly from a tarball.

goneri:
The problem are the symlinks and the situation is unlikely to be fixed any-time soon.

pabelanger:
Can you point me to the issue with symlinks?  Because, I feel AH is also going to have the problem, which will need to be fixed regardless.

goneri:
It'ss the invalid-documentation errors in my example above. These modules are actually symlinks: https://github.com/ansible-collections/splunk.es/tree/master/plugins/modules

```
$ cd /home/goneri/.ansible/collections/ansible_collections/splunk/es
$ ansible-test sanity --test validate-modules 2>&1|grep DOCUMENTATION.module
# No error
```

Now, if I replace the symlink with the file content:
```
$ cd /home/goneri/.ansible/collections/ansible_collections/splunk/es
$ rm plugins/modules/splunk_adaptive_response_notable_event.py
$ cp plugins/modules/adaptive_response_notable_event.py plugins/modules/splunk_adaptive_response_notable_event.py
$ ansible-test sanity --test validate-modules 2>&1|grep DOCUMENTATION.module
$ ERROR: plugins/modules/splunk_adaptive_response_notable_event.py:0:0: invalid-documentation: DOCUMENTATION.module: not a valid value for dictionary value @ data['module']. Got 'adaptive_response_notable_event'

```

pabelanger:
Just to follow up here, in the case of symlinks I think it is best, to just ignore the errors from that file for now, until ansible-test / ansible-galaxy is fixed.

Can use something like: https://github.com/ansible-collections/vyos.vyos/blob/master/tests/sanity/ignore-2.9.txt

goneri:
Superseded by https://github.com/ansible-collections/splunk.es/pull/4, as suggested by Paul.

PR #529, opened 2020-06-15
avoid the jq/yq dependency

goneri:
Replace `yq` calls by `include_vars` from core.

### DIFF
--- a/playbooks/ansible-test-base/run.yaml
+++ b/playbooks/ansible-test-base/run.yaml
@@ -1,10 +1,21 @@
 ---
 - hosts: controller
   tasks:
-    - name: Fetch and install the artifacts
-      import_role:
-        name: deploy-artifacts
-      when: ansible_test_collections | default(False)
+    - when: ansible_test_collections | default(False)
+      block:
+        - name: Fetch and install the artifacts
+          import_role:
+            name: deploy-artifacts
+        - name: Copy the galaxy.yml on the controller
+          fetch:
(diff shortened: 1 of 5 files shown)


playbooks/ansible-test-base/run.yaml
```diff
@@ -1,10 +1,21 @@
...
-    - name: Fetch and install the artifacts
-      import_role:
-        name: deploy-artifacts
-      when: ansible_test_collections | default(False)
+    - when: ansible_test_collections | default(False)
+      block:
+        - name: Fetch and install the artifacts
+          import_role:
+            name: deploy-artifacts
+        - name: Copy the galaxy.yml on the controller
```
pabelanger, line 9: Can't this be simplified using slurp?

goneri: Not really, `slurp` stores in memory, `include_vars` reads from a file.

pabelanger:
we pip install yq in ansible-test-base I think, lets also drop it

goneri:
> we pip install yq in ansible-test-base I think, lets also drop it

Hum, where?

pabelanger:
> > we pip install yq in ansible-test-base I think, lets also drop it
> 
> Hum, where?

Ah, my mistake this was deleted in the PR. Okay, merge when ready

goneri:
Merging, I will keep an eye on the results today.

PR #542, opened 2020-06-17
add ansible-test-sanity-qradar-stable29

goneri:


zuul.d/ansible-security-jobs.yaml
```diff
@@ -104,3 +104,10 @@
     timeout: 3600
     vars:
       ansible_collections_repo: github.com/ansible-collections/ibm.qradar
+
+- job:
+    name: ansible-test-sanity-qradar-stable29
+    parent: ansible-test-sanity-qradar
```
pabelanger, line 110: sanity is already 2.9, so this is a noop job

PR #545, opened 2020-06-17
run some single node jobs on AWS

goneri:
The goal of this job is to validate we can run single node jobs on AWS.

### DIFF
--- a/zuul.d/jobs.yaml
+++ b/zuul.d/jobs.yaml
@@ -19,17 +19,12 @@
 - job:
     name: ansible-tox-docs
     parent: tox-docs
-    nodeset: centos-8-1vcpu
+    nodeset: centos-8-1vcpu-aws
 
 - job:
     name: ansible-tox-linters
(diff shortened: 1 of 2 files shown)


goneri:
recheck

goneri:
recheck

goneri:
recheck

goneri:
recheck

goneri:
recheck

goneri:
recheck

goneri:
recheck

pabelanger:
> The goal of this job is to validate we can run single node jobs on AWS.

I don't mind we we add nodes into aws, but we cannot limit jobs just to aws.  This would be a step backwards for jobs to be run only in a single region. And, we won't be able to gate jobs with that, for fear of an outage will block zuul from merging code.

goneri:
Superseded by https://github.com/ansible-network/windmill-config/pull/696

PR #549, opened 2020-06-18
Don't install ara for ansible-test unit

pabelanger:
There is no need, as we don't use ara to render results.


ganeshrn:
Unit test failures are related to PR https://github.com/ansible/ansible-zuul-jobs/pull/549 and should be fixed after the PR is merged

PR #576, opened 2020-06-26
Fix `sphinx_build_dir` value for ansible-lint

webknjaz:
Ref: https://github.com/ansible/ansible-lint/pull/858


webknjaz:
@pabelanger ^

PR #578, opened 2020-06-27
ansible-lint: correct docs html location

ssbarnea:
Change sphinx_build_dir to a folder that will contain the expected html subdirectory.

Needed-By: https://github.com/ansible/ansible-lint/pull/865

### DIFF
--- a/zuul.d/jobs-ansible-lint.yaml
+++ b/zuul.d/jobs-ansible-lint.yaml
@@ -5,7 +5,8 @@
     name: ansible-lint-tox-docs
     parent: ansible-tox-docs
     vars:
-      sphinx_build_dir: .tox/docs_html
+      # zuul expects a `html` folder to be created under this
+      sphinx_build_dir: .tox/docs
 
 - job:
     name: ansible-lint-tox-linters


ssbarnea:
@webknjaz

ssbarnea:
It would be incorrect to say conflict because the output is expected to be contained inside the same temp directory used by tox. This assures that when you do “tox -e docs” it will also remove the previous output.

If you are concerned about possible clashing it can add .tox/docs/logs — so the produced output will be inside logs folder. I used this approach in other cases because zuul knows to collect everything from inside that folder without extra changes.

Note: every directory under .tox should be mapped to a tox env, not something else.

zuul.d/jobs-ansible-lint.yaml
```diff
@@ -5,7 +5,8 @@
     name: ansible-lint-tox-docs
     parent: ansible-tox-docs
     vars:
-      sphinx_build_dir: .tox/docs_html
+      # zuul expects a `html` folder to be created under this
+      sphinx_build_dir: .tox/docs
```
webknjaz, line 9:
```suggestion
      sphinx_build_dir: .tox/docs/html
```
(as discussed on freenode)

webknjaz:
@pabelanger ^

PR #583, opened 2020-07-01
Pulp: Introduce installer jobs

Spredzy:
Depends-On: https://github.com/ansible/project-config/pull/534

### DIFF
--- a/playbooks/pulp/pre.yaml
+++ b/playbooks/pulp/pre.yaml
@@ -1,0 +1,22 @@
+---
+- hosts: all
+  tasks:
+    - name: Ensure git is installed
+      package:
+        name: git
+      become: true
+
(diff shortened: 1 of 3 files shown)


playbooks/pulp/pre.yaml
```diff
@@ -0,0 +1,22 @@
+---
+- hosts: all
+  tasks:
+    - name: Ensure git is installed
```
Spredzy, line 4: @pabelanger Can we assume this is always the case ?

pabelanger: no, it is safer to add it into bindep.txt file in your project repo, then we'll ensure it is installed

Spredzy:
https://github.com/ansible/project-config/pull/534#issuecomment-652956021

PR #588, opened 2020-07-07
Avoid creating swap by default

ssbarnea:
Avoids outage with fedora-31 since yesterday and makes our jobs
more resilient. Swap existence is a customization, which should
not be enabled by default.

### DIFF
--- a/playbooks/base/pre.yaml
+++ b/playbooks/base/pre.yaml
@@ -18,6 +18,12 @@
 - hosts: all:!appliance*
   tasks:
     - name: Run configure-swap role
+      # configuring swap can fail for multiple reasons and should not be
+      # enabled unless requested
+      # https://bugs.launchpad.net/ubuntu/+source/util-linux/+bug/1788321
+      # https://unix.stackexchange.com/questions/294600/i-cant-enable-swap-space-on-centos-7
+      # https://bugzilla.redhat.com/show_bug.cgi?id=1129205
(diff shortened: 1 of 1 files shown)


ssbarnea:
@goneri @pabelanger please check. It should address https://dashboard.zuul.ansible.com/t/ansible/builds?job_name=ansible-tox-py37

pabelanger:
Thank you for looking, but we always want a swap. We should fix the issue of why this is failing, not disable it.

playbooks/base/pre.yaml
```diff
@@ -18,6 +18,12 @@
 - hosts: all:!appliance*
   tasks:
     - name: Run configure-swap role
+      # configuring swap can fail for multiple reasons and should not be
+      # enabled unless requested
+      # https://bugs.launchpad.net/ubuntu/+source/util-linux/+bug/1788321
+      # https://unix.stackexchange.com/questions/294600/i-cant-enable-swap-space-on-centos-7
+      # https://bugzilla.redhat.com/show_bug.cgi?id=1129205
+      when: configure_wrap | default(false)
```
goneri, line 26: `configure_wrap` -> `configure_swap` :-)

- I would call it `enable_swap` to avoid the name confusion with the role name
- Since it's the default today for all our jobs, the value has to default to `True`, to avoid any disruption.

pabelanger:
we can close this

PR #600, opened 2020-07-15
network: move py{35,37,38} jobs in periodical pipeline

goneri:
There is currently quite a few running jobs but they consume wy too much jobs
and as a result, we are often at capacity. Our resources are not unlimited,
and in this case, jobs just stay pending... for hours.

For instance, we've got currently a vmware job that has been running
for 16h now.

This kind of delay does not benefit anyone, they impact our moral,
the contributors motivations and they certainly don't improve the quality
of the collections.

Python35, Python37 and Python38 are not criticial and consume a lot of
resources. They can definitiely be run in a delayed pipeline. And this would
help use greatly to speed up our CI.

### DIFF
--- a/zuul.d/project-templates.yaml
+++ b/zuul.d/project-templates.yaml
@@ -78,29 +78,38 @@
         - ansible-test-network-integration-eos-httpapi-python27
         - ansible-test-network-integration-eos-httpapi-python27-stable210
         - ansible-test-network-integration-eos-httpapi-python27-stable29
-        - ansible-test-network-integration-eos-httpapi-python35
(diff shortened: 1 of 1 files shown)


pabelanger:
While I agree, we don't want to move these into periodic as no one will look at the results I fear.  A better idea is doing something like https://github.com/ansible/ansible-zuul-jobs/pull/597 which we are committed too.

I think we could move arista.eos and vyos.vyos today, then move other collections as each coverage report is improved.

goneri:
https://github.com/ansible/ansible-zuul-jobs/pull/601 should improve the situation

PR #610, opened 2020-07-20
Update bootstrap.yaml

Qalthos:
This should unblock httpapi, but I'm not sure if it's the best solution


pabelanger:
https://github.com/ansible/ansible-zuul-jobs/pull/611 is actually the fix here, we are not running the 2nd bootstrap playbook properly any more. This was due to an upstream change to zuul-jobs that broke us.

PR #629, opened 2020-08-21
ansible-test: add molecule test

goneri:
Depends-On: https://github.com/ansible/ansible-zuul-jobs/pull/627

Add two molecule scenarios for ansible-test role. Both are based on podman and target the `sanity` sub-command.


goneri:
@ssbarnea what would be the best option to run these scenarios in the CI?

PR #653, opened 2020-09-22
DNM: inject my key for root access

GomathiselviS:


justjais:
@GomathiselviS @pabelanger can we get this merged as it has the depends-on for the PR: https://github.com/ansible-collections/trendmicro.deepsec/pull/7, of if NR can I remove the depends-on tag from respective PR.

PR #655, opened 2020-09-23
Removing ara installation

GomathiselviS:
Ara installs ansible2.10, though we want ansible 2.9 . Removing installation of ara, for now, to unblock CI.

### DIFF
--- a/playbooks/ansible-test-base/pre.yaml
+++ b/playbooks/ansible-test-base/pre.yaml
@@ -28,9 +28,10 @@
       shell: ~/venv/bin/pip install selinux "setuptools<50.0.0"
       when: ansible_os_family == "RedHat"
 
-    - name: Install ara into virtualenv
-      shell: ~/venv/bin/pip install "ara<1.0.0"
-      when: "'integration' in ansible_test_command"
+    # ara is installing ansible-2.10, when 2.9 is needed. Commenting to temporarily fix the issue.
+    # - name: Install ara into virtualenv
+    #   shell: ~/venv/bin/pip install "ara<1.0.0"
+    #   when: "'integration' in ansible_test_command"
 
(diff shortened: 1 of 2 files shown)


playbooks/ansible-test-base/pre.yaml
```diff
@@ -28,9 +28,10 @@
       shell: ~/venv/bin/pip install selinux "setuptools<50.0.0"
       when: ansible_os_family == "RedHat"
 
-    - name: Install ara into virtualenv
-      shell: ~/venv/bin/pip install "ara<1.0.0"
-      when: "'integration' in ansible_test_command"
+    # ara is installing ansible-2.10, when 2.9 is needed. Commenting to temporarily fix the issue.
+    # - name: Install ara into virtualenv
+    #   shell: ~/venv/bin/pip install "ara<1.0.0"
+    #   when: "'integration' in ansible_test_command"
```
dmsimard, line 34: FYI this is because ara 0.x pulls in Ansible to load it's configuration -- it certainly doesn't have a requirement on Ansible 2.10, it's just that the requirement is unpinned so it picks up whatever is the latest version.

Another fix would be to pin Ansible to <2.10.

Also, out of scope for now but also note that ara 1.x no longer has a dependency on Ansible :)

goneri: Agreed, we should jump to a new version Ara.

GomathiselviS: I can try the new version of Ara, in a separate PR. I feel that this fix will currently unblock CI.

goneri: @Gomathiselvi-S I would like to see this patch merged too, to fix the CI ASAP.

PR #670, opened 2020-10-07
Keep ios fail fast only for gate job and remove for check job

justjais:
Keep ios fail fast only for gate job and remove for check job

### DIFF
--- a/zuul.d/project-templates.yaml
+++ b/zuul.d/project-templates.yaml
@@ -212,7 +212,6 @@
 - project-template:
     name: ansible-collections-cisco-ios-units
     check:
-      fail-fast: true
       jobs: &ansible-collections-cisco-ios-units-jobs
         - ansible-changelog-fragment
         - ansible-test-sanity-ios


pabelanger:
we don't want to do this, we should keep both gate and check the same. The reasons is, gate is a dependent pipeline which can affect other projects PRs, creating gate resets.

What we really should do, is audit the failures and work to fix them, not ignore them. Thoughts?

justjais:
@pabelanger I had removed the check job fail fast coz it was failing the test run at the encounter of 1st failure which was resulting into back n forth fixing of check test jobs and that's why applied the fail fast only to gate jobs as that'd fail mostly for non-tech issues if the check jobs have passed.

Also I didn't get the idea you were trying to convey here  _(What we really should do, is audit the failures and work to fix them, not ignore them. Thoughts?)_

pabelanger:
the main goal is to keep check and gate in alignment, because if we hit the issue in check, we are also likely to hit it in gate.  But yes, what you are saying is true.  It makes it harder to debug multiple failures.

Lets trying, and adjust if needed.

PR #671, opened 2020-10-07
Enabled voting for vyos jobs

GomathiselviS:

Now that all the vyos jobs are passing, voting is enabled on them.

### DIFF
--- a/zuul.d/project-templates.yaml
+++ b/zuul.d/project-templates.yaml
@@ -594,7 +594,7 @@
 - project-template:
     name: ansible-collections-vyos-vyos
     check:
-      jobs:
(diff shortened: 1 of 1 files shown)


zuul.d/project-templates.yaml
```diff
@@ -595,21 +595,15 @@
     name: ansible-collections-vyos-vyos
     check:
       jobs:
```
pabelanger, line 597: we should add an yaml anchor, so we can also ensure the same jobs are in gate.  As it stands, gate do not have local jobs now

GomathiselviS:
recheck

PR #673, opened 2020-10-07
sanity: default val for ansible_collections_repo

goneri:
Depends-On: https://github.com/ansible/ansible-zuul-jobs/pull/672

Iniialize `ansible_collections_repo` with `zuul.project.canonical_name`.
This should actually suites all the cases.


goneri:
See: https://github.com/ansible/ansible-zuul-jobs/pull/674

PR #674, opened 2020-10-07
vmware: use the generic jobs

goneri:
Use `ansible-test-sanity-base-210` directly.


goneri:
@pabelanger any objection if I do something like that?

PR #679, opened 2020-10-21
Ignore integration tests on docs-only PRs

Qalthos:


zuul.d/jobs.yaml
```diff
@@ -308,8 +308,10 @@
       - name: github.com/ansible/ansible
       - name: github.com/ansible-collections/cisco.asa
     timeout: 9000
-    irrelevant-files:
+    irrelevant-files: &documentation-files
```
pabelanger, line 311: I think we could move this into ansible-network-appliance-base job to cut down on the copy / pasta

PR #684, opened 2020-10-26
test cloud.common with vmware.vmware_rest

goneri:
Run `ansible-test-cloud-integration-vmware-rest-python36` to validate
`cloud.common`


goneri:
recheck

PR #693, opened 2020-10-29
Drop all local tests except vyos

Qalthos:
This should be sufficient coverage for local

### DIFF
--- a/zuul.d/jobs.yaml
+++ b/zuul.d/jobs.yaml
@@ -1978,6 +1978,13 @@
       ansible_test_network_cli_ssh_type: libssh
 
 - job:
+    name: ansible-test-network-integration-vyos-local-python36-stable210
+    parent: ansible-test-network-integration-vyos-local-python36
+    required-projects:
+      - name: github.com/ansible/ansible
+        override-checkout: stable-2.10
+
+- job:
     name: ansible-test-network-integration-vyos-network_cli-python36-stable210
     parent: ansible-test-network-integration-vyos-network_cli
     nodeset: vyos-1.1.8-python36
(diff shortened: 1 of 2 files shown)


zuul.d/jobs.yaml
```diff
@@ -2072,16 +1977,6 @@
     vars:
       ansible_test_network_cli_ssh_type: libssh
 
-- job:
```
pabelanger, line 2075: Shouldn't we keep testing local on stable branches?  I think for vyos, we should be safe

zuul.d/jobs.yaml
```diff
@@ -2098,16 +1993,6 @@
...
```
pabelanger, line 2101: same

zuul.d/project-templates.yaml
```diff
@@ -597,13 +592,9 @@
     name: ansible-collections-vyos-vyos
     check:
       jobs: &ansible-collections-vyos-vyos-jobs
-        - ansible-test-network-integration-vyos-local-python27-stable210
```
pabelanger, line 600: I think we should keep stable branches

Qalthos: Did you mean this to be on 2.7, or the 3.6 runs below?

pabelanger: yah, we still want to support python27 and local on stable branches. We've only dropped py27 on devel right now

Qalthos: So... are we not dropping any local tests on vyos then?

pabelanger: yah, I think we need to keep all on vyos, but can drop them every place else.

zuul.d/jobs.yaml
```diff
@@ -1991,13 +1929,6 @@
     vars:
       ansible_test_python: 3.5
 
-- job:
```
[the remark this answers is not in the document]
Qalthos:
```
% git grep "vyos-local-python27"
zuul.d/jobs.yaml:    name: ansible-test-network-integration-vyos-local-python27-stable210
zuul.d/jobs.yaml:    name: ansible-test-network-integration-vyos-local-python27-stable29
zuul.d/project-templates.yaml:        - ansible-test-network-integration-vyos-local-python27-stable210
zuul.d/project-templates.yaml:        - ansible-test-network-integration-vyos-local-python27-stable29
```
As you said, we stopped running 2.7 on devel, so why keep the job around? It's not even a parent for the stable jobs.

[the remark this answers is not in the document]
pabelanger: oh, you are right. This is left over when I removed it yesterday

PR #767, opened 2021-02-01
To enable tm integration tests job over aws instance

justjais:
To enable tm integration tests job over aws instance

### DIFF
--- a/zuul.d/ansible-security-jobs.yaml
+++ b/zuul.d/ansible-security-jobs.yaml
@@ -151,3 +151,9 @@
     timeout: 3600
     vars:
       ansible_collections_repo: github.com/ansible-collections/trendmicro.deepsec
+
+- job:
+    name: ansible-security-integration-trendmicro-deepsec-python36
+    parent: ansible-security-integration-trendmicro-deepsec
+    vars:
+      ansible_test_python: 3.6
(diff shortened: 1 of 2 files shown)


zuul.d/ansible-security-jobs.yaml
```diff
@@ -151,3 +151,9 @@
     timeout: 3600
     vars:
       ansible_collections_repo: github.com/ansible-collections/trendmicro.deepsec
+
+- job:
+    name: ansible-security-integration-trendmicro-deepsec-python36
+    parent: ansible-security-integration-trendmicro-deepsec
```
GomathiselviS, line 157:
```suggestion
    parent: ansible-test-security-integration-trendmicro-deepsec
```

zuul.d/ansible-security-jobs.yaml
```diff
@@ -151,3 +151,9 @@
     timeout: 3600
     vars:
       ansible_collections_repo: github.com/ansible-collections/trendmicro.deepsec
+
+- job:
```
pabelanger, line 155: This is missing a nodeset.

The way to test this, would be to open a PR in trendmicro collection, then depends-on this PR. That would run your job

PR #780, opened 2021-02-11
Refactor ansible-test-sanity jobs

pabelanger:
This moves all collections to use ansible-test-sanity-docker, to help
keep jobs supported in ansible community package.

### DIFF
--- a/zuul.d/ansible-community-jobs.yaml
+++ b/zuul.d/ansible-community-jobs.yaml
@@ -1,9 +1,0 @@
----
-- job:
-    name: ansible-test-sanity-community-asa
-    parent: ansible-test-sanity-base-29
-    required-projects:
-      - name: github.com/ansible-collections/community.asa
-    timeout: 3600
-    vars:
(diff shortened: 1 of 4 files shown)


goneri:
I would really like if we can we replace the name `docker` with something more agnostic like `container`.

pabelanger:
> I would really like if we can we replace the name `docker` with something more agnostic like `container`.

I'd usually agree, the purpose here is to help show we are using docker in ansible-test, given it only supports docker.  I do think we should push ansible-test to support any oci version, if I am honest.

goneri:
`ansible-test` (devel) works just fine with Podman thinks to a small patch. I just pushed backport PR for review with this fix:

- https://github.com/ansible/ansible/pull/73570
- https://github.com/ansible/ansible/pull/73569

pabelanger:
For now, I'm going to land this. If we want to refactor into podman or have ansible-test be more generic, I think that is great. We have some pressure to do this now, and mostly why we pushing on it.

PR #787, opened 2021-02-19
Revert "Revert "Remove vcenter7 jobs""

pabelanger:
Sorry, we are going to revert this again until we can debug nodepool.
Something is not right with vmware, and in the process we end up leaking
65 nodes in a ready state.

This reverts commit 9a9b4034cee446a6a4ee21025f95a00d188dbe93.


pabelanger:
@goneri Sorry, reverting again. I'd like to sync with you tomorrow to see how we can properly debug this.  Basically, these jobs are consuming all our quota if we cannot allocate all the nodes in the nodeset the first time.

We may have to use a semaphore to limit the number of concurrent jobs we run here.

PR #791, opened 2021-02-23
rename  ansible-test-provider

goneri:
Depends-On: https://github.com/ansible/ansible-zuul-jobs/pull/796

Rename `vmware-ci-configure-ansible-test` to `ansible-test-provider` in
order to be able to use it with aws.


goneri:
recheck

PR #793, opened 2021-02-25
Make cisco.nxos jobs voting

NilashishC:


pabelanger:
So, we don't want to do this right now. As nxos images, are kinda of ugly and large.  I think the order is the following:

1. upgrade nxos images to latest version
2. Work to see if all providers support UTFI bios, today we do nested virt with nxos which require a lot of quota
3. We really need to discuss the amount of quota these images use, this is why we run into NODE_FAILUREs and the main reason these jobs are non-voting.

lastthyme (approved):

PR #797, opened 2021-03-01
vmware: move specific tasks in single playbook

goneri:
Depends-On: https://github.com/ansible/ansible-zuul-jobs/pull/799

Move vcenter specific tasks from playbooks/ansible-test-cloud-integration-base/pre.yaml to playbooks/ansible-cloud/vcenter-appliance/pre.yaml


goneri:
recheck

PR #817, opened 2021-03-05
FIPS Testing with eos-appliance

GomathiselviS:

Depends-On: https://github.com/ansible/ansible/pull/74170
Depends-On: https://github.com/ansible/ansible/pull/74169

FIPS testing for eos


pabelanger:
recheck

playbooks/ansible-test-network-integration-base/files/bootstrap-eos.yaml
```diff
@@ -1,8 +1,20 @@
 ---
 - hosts: appliance
-  gather_facts: false
   connection: network_cli
   tasks:
+    - name: lookup SSH public key (FIPS)
+      set_fact:
+        _ssh_key: "{{ lookup('file', '~/.ssh/id_ecdsa.pub') }}"
+      when: "'libssh' in zuul.job"
```
pabelanger, line 8: rather then checking for libssh, we could just default to ecdsa keys for eos now

playbooks/ansible-test-network-integration-base/files/bootstrap-eos.yaml
```diff
@@ -1,8 +1,20 @@
...
+
+    - name: Setup zuul ssh key (FIPS)
+      become: true
+      eos_config:
+        lines:
+          - "username zuul sshkey {{ _ssh_key.split()[0] }} {{ _ssh_key.split()[1] }}"
+        save_when: always
+      when: "'libssh' in zuul.job"
```
pabelanger, line 16: same comment, lets just default to ecdsa keys

playbooks/ansible-test-base/run.yaml
```diff
@@ -13,6 +13,14 @@
             file: '{{ _fetch.dest }}'
             name: galaxy_info
 
+    - name: Enable FIPS mode
+      when:
+        - ansible_distribution == "CentOS"
+        - hostvars[item].ansible_network_os in ['eos']
```
pabelanger, line 19: This means all eos appliances will have FIPs enabled, this is a problem for paramiko. Since it won't support it.

We likely need an network_os check, maybe a denylist by default, along with libssh check

GomathiselviS:
recheck

GomathiselviS:
recheck

roles/enable-fips-mode/tasks/main.yaml
```diff
@@ -0,0 +1,25 @@
...
+  reboot:
+    reboot_timeout: 1800
+
+- name: Run start-zuul-console role
+  include_role:
+    name: start-zuul-console
+
+- name: Ensure FIPS mode is enabled
+  become: true
+  shell: fips-mode-setup --check
```
pabelanger, line 25: We likely should capture the output and validate that fips mode is enabled, and fail if not. That way, we can debug fips failures, if they ever happen.

playbooks/ansible-test-base/run.yaml
```diff
@@ -13,6 +13,15 @@
             file: '{{ _fetch.dest }}'
             name: galaxy_info
 
+    - name: Enable FIPS mode
+      when:
+        - ansible_distribution == "CentOS"
+        - ansible_test_network_cli_ssh_type == "libssh"
+        - hostvars[item].ansible_network_os in ['eos']
```
pabelanger, line 20: Rather then network_os check, I wonder if we should add a new zuul var to the base appliance job call fipsmode: True, then jobs will disable it if they don't want it use it.

Thoughts?

GomathiselviS: Sounds reasonable. I will add it.

PR #820, opened 2021-03-11
To add TM Deepsecurity silent configuration setup role

justjais:
To add TM Deepsecurity silent configuration setup role

PR #825, opened 2021-03-18
Changes to update to TM tests setup play to use newly built  TM role

justjais:
Depends-On: https://github.com/ansible/ansible-zuul-jobs/pull/820

Changes to update to TM tests setup play to use newly built  TM role for making work the silent setup for TM over AWS instance.


goneri:
recheck

zuul.d/project.yaml
```diff
@@ -125,6 +125,7 @@
         - ansible-security-trendmicro-deepsec-appliance:
             files:
               - playbooks/ansible-network-appliance-base/.*
+              - roles/tm.*
```
goneri, line 128: You can remove the files section for your test. This way `ansible-security-tredmicro-deepsec-appliance` will be always executed, even if there is no changes in these directories

zuul.d/project.yaml
```diff
@@ -122,9 +122,7 @@
               - playbooks/ansible-test-network-integration-base/files/bootstrap-splunk.yaml
               - playbooks/ansible-network-appliance-base/.*
               - roles/splunk.*
-        - ansible-security-trendmicro-deepsec-appliance:
-            files:
-              - playbooks/ansible-network-appliance-base/.*
+        - ansible-test-security-integration-trendmicro-deepsec-python38
```
goneri, line 125: You can also comment out the vmware and openvswitch jobs. Just to save some resources.

goneri: You did the change the `gate` section, you should actually add this line in the `check` part, e.g around line 85..

justjais:
closing the PR as have transferred all of the content to PR #820

PR #830, opened 2021-03-23
Add test sanity base job for ansible 2.11 and enable it for iosxr

ashwini-mhatre:


zuul.d/jobs.yaml
```diff
@@ -164,6 +164,27 @@
...
+    parent: unittests
+    nodeset: controller-python36
+    dependencies:
+      - name: build-ansible-collection
+        soft: true
+    pre-run: playbooks/ansible-test-base/pre.yaml
+    run: playbooks/ansible-test-base/run.yaml
+    required-projects:
+      - name: github.com/ansible/ansible
+        override-checkout: devel
```
pabelanger, line 178: lets add
```
#NOTE(username): Switch to stable-2.11 once branch is created.
```

Something like: https://github.com/ansible/ansible-zuul-jobs/blob/master/zuul.d/jobs.yaml#L260

zuul.d/jobs.yaml
```diff
@@ -164,6 +164,28 @@
...
+    parent: unittests
+    nodeset: controller-python36
+    dependencies:
+      - name: build-ansible-collection
+        soft: true
+    pre-run: playbooks/ansible-test-base/pre.yaml
+    run: playbooks/ansible-test-base/run.yaml
+    required-projects:
+      - name: github.com/ansible/ansible
+        # NOTE (amhatre): Switch to stable-2.11 once branch is created.
```
pabelanger, line 178: You could remove the space between # and (

Not sure if linters will fail

PR #845, opened 2021-03-29
Add ansible-navigator smoke test job

relrod:
Adds a job that can run `tox -e smoke` for running `ansible-navigator` smoke tests...

### DIFF
--- a/zuul.d/ansible-navigator-jobs.yaml
+++ b/zuul.d/ansible-navigator-jobs.yaml
@@ -1,0 +1,7 @@
+---
+- job:
+    name: ansible-navigator-tox-smoke
+    parent: tox
+    vars:
+      tox_envlist: smoke
+    nodeset: centos-8-1vcpu
--- a/zuul.d/jobs.yaml
+++ b/zuul.d/jobs.yaml
@@ -463,10 +463,3 @@
     vars:
       container_command: podman
     nodeset: ubuntu-bionic-2vcpu
-
-- job:
-    name: ansible-navigator-tox-smoke
-    parent: tox
-    vars:
-      tox_envlist: smoke
-    nodeset: centos-8-1vcpu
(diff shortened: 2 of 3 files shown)


zuul.d/jobs.yaml
```diff
@@ -463,3 +463,10 @@
     vars:
       container_command: podman
     nodeset: ubuntu-bionic-2vcpu
+
```
pabelanger, line 466: This looks right, but lets move it into zuul.d/ansible-navigator-jobs.yaml

pabelanger: You can also add the job to https://github.com/ansible/ansible-zuul-jobs/blob/master/zuul.d/project-templates.yaml#L54 which attaches it to the check / gate pipelines for the navigator project.

cidrblock: thx paul

pabelanger:
@cidrblock I would actually revert this until you get the job properly working. There is no need to merge to get zuul to use speculative job configuration.

zuul.d/ansible-navigator-jobs.yaml
```diff
@@ -0,0 +1,7 @@
+---
+- job:
+    name: ansible-navigator-tox-smoke
+    parent: tox
```
pabelanger, line 4: it is possible you want to use ansible-tox-py38 here, over tox

pabelanger: this will allow you to use bindep.txt files for installs

relrod: Is there anything special I have to do for that? Just switch it here and then drop a bindep.txt in the root of the other repo?

relrod: Opened #849 to change it, but actually if I'm reading right, `tox` should be enough? https://opendev.org/zuul/zuul-jobs/src/branch/master/zuul.d/python-jobs.yaml#L36

PR #847, opened 2021-03-30
update override_checkout variable from devel to stable-2.11

ashwini-mhatre:


relrod:
(Note that 2.11rc1 got delayed a week, so it hasn't branched yet. Current plan is to branch this coming Monday, 5 April.)

ashwini-mhatre:
> (Note that 2.11rc1 got delayed a week, so it hasn't branched yet. Current plan is to branch this coming Monday, 5 April.)

yeah we can merge this PR after that.shall i add DNM for now?

PR #850, opened 2021-03-31
vmare: uses nested virt everywhere

goneri:
Since vSphere 7.0.1, vSphere associates a VM to all the DVS switch.
As a result it becomes really difficult to identify the scenarios
that start a VM with the others.

This is a first step toward breaking the segregation between the
two class of tests.


goneri:
supereseded by https://github.com/ansible/ansible-zuul-jobs/pull/843

PR #853, opened 2021-04-01
Change ssh key to ECDSA  from RSA for eos appliance

GomathiselviS:

This change is made to support FIPS mode.

### DIFF
--- a/zuul.d/arista-eos-jobs.yaml
+++ b/zuul.d/arista-eos-jobs.yaml
@@ -6,9 +6,6 @@
     parent: ansible-network-appliance-base
     pre-run: playbooks/ansible-network-eos-appliance/pre.yaml
     run: playbooks/ansible-network-eos-appliance/run.yaml
-    vars:
-      zuul_ssh_key_algorithm: ecdsa
-      zuul_ssh_key_size: 521
     host-vars:
       eos-4.20.10:
         ansible_connection: network_cli
--- a/zuul.d/arista-eos-jobs.yaml
+++ b/zuul.d/arista-eos-jobs.yaml
@@ -6,6 +6,9 @@
     parent: ansible-network-appliance-base
     pre-run: playbooks/ansible-network-eos-appliance/pre.yaml
     run: playbooks/ansible-network-eos-appliance/run.yaml
+    vars:
+      zuul_ssh_key_algorithm: ecdsa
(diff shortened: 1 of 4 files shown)


zuul.d/arista-eos-jobs.yaml
```diff
@@ -6,6 +6,9 @@
     parent: ansible-network-appliance-base
     pre-run: playbooks/ansible-network-eos-appliance/pre.yaml
     run: playbooks/ansible-network-eos-appliance/run.yaml
+    vars:
+      zuul_ssh_key_algorithm: ecdsa
+      zuul_ssh_key_size: 521
```
pabelanger, line 11: We won't do this any more. The fix will be to update base-minimal-test to generate both rsa / ecdsa keys so both keys are on disk

pabelanger: https://github.com/ansible/project-config/blob/master/playbooks/base-minimal-test/pre.yaml#L15

if you want to propose a PR to remove our fork, that is step 1

GomathiselviS:
recheck

zuul.d/project.yaml
```diff
@@ -8,6 +8,7 @@
             files:
               - playbooks/ansible-network-appliance-base/.*
               - playbooks/ansible-network-asa-appliance/.*
+            voting: false
```
pabelanger, line 11: if you revert this, the jobs won't run

zuul.d/project.yaml
```diff
@@ -51,11 +52,6 @@
               - playbooks/ansible-test-network-integration-base/files/bootstrap-splunk.yaml
               - playbooks/ansible-network-appliance-base/.*
               - roles/splunk.*
-        - ansible-security-trendmicro-deepsec-appliance:
-            files:
```
pabelanger, line 55: same

zuul.d/project.yaml
```diff
@@ -74,6 +70,7 @@
             files:
               - playbooks/ansible-network-appliance-base/.*
               - playbooks/ansible-network-asa-appliance/.*
+            voting: false
```
pabelanger, line 73: same

pabelanger:
lets revert your changes, to zuul config. The jobs won't run, now because your base-minimal-test change is gone.

PR #857, opened 2021-04-05
Use anchors for common data

Akasurde:
Instead of repeating jobs use anchors

### DIFF
--- a/zuul.d/project-templates.yaml
+++ b/zuul.d/project-templates.yaml
@@ -41,13 +41,11 @@
       Runs unit tests for an Ansible Python project under the CPython
       version 3 releases.
     check:
-      jobs:
+      jobs: &ansible-python3-jobs
         - ansible-tox-py36
         - ansible-tox-py37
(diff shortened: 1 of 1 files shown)


zuul.d/project-templates.yaml
```diff
@@ -102,23 +90,12 @@
...
-        - ansible-test-sanity-docker-stable-2.10
-        - ansible-test-units-community-aws-python38
-        - build-ansible-collection:
-            required-projects:
-              - name: github.com/ansible-collections/amazon.aws
-              - name: github.com/ansible-collections/ansible.utils
-              - name: github.com/ansible-collections/community.general
-        - ansible-test-cloud-integration-aws-py36:
-            voting: false
+      jobs: *ansible-collections-community-aws-jobs
```
pabelanger, line 93: This isn't right, we shouldn't have non-voting jobs in gate.  The fix here, is to remove that job, over anchor

PR #874, opened 2021-04-22
Add semaphore to ansible-test-cloud-integration-vmware-rest

pabelanger:
Due to limited capacity in vexxhost, it is possible for this job to
wedge vexxhost. Given how much capacity it uses.

### DIFF
--- a/zuul.d/ansible-cloud-jobs.yaml
+++ b/zuul.d/ansible-cloud-jobs.yaml
@@ -235,6 +235,12 @@
         ansible_network_os: vmware_rest
     nodeset: vmware-vcsa_1esxi-7.0.2-python36-with-nested
 
+# NOTE(pabelanger): Due to low memory capacity in vexxhost, it is
+# possible to wedge the provider.  For now, only run 1 job at a time.
+- semaphore:
+    name: ansible-test-cloud-integration-vmware-rest
+    max: 1
+
 - job:
     name: ansible-test-cloud-integration-vmware-rest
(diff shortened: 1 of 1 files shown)


zuul.d/ansible-cloud-jobs.yaml
```diff
@@ -235,6 +235,12 @@
         ansible_network_os: vmware_rest
     nodeset: vmware-vcsa_1esxi-7.0.2-python36-with-nested
 
+# NOTE(pabelanger): Due to low memory capacity in vexxhost, it is
+# possible to wedge the provider.  For now, only run 1 job at a time.
+- semaphore:
+    name: ansible-test-cloud-integration-vmware-rest
+    max: 1
```
goneri, line 242: @pabelanger this is way to low. Right now there is just 13 Vexxhost instances in-use and the VMware jobs are all stuck for hours.

pabelanger: Yes, this means only 1 concurrent job at at time; lets bump to 2.

The other issue, is the esxi jobs which are multiple nodes, conflict here.

Basically, because we only run things on vexxhost now, we are severely capped for vmware jobs.

Limestone is twice as big.

pabelanger: why are we not launching vmware in sjc1 for vexxhost?

goneri: This is weird, we should. I will take a look tomorrow.

PR #885, opened 2021-04-29
Replace stable210 with stable211 jobs for eos

GomathiselviS:

Dropping stable-210 and adding stable-211 jobs for eos.


zuul.d/project-templates.yaml
```diff
@@ -120,7 +120,6 @@
         - ansible-changelog-fragment
         - ansible-test-sanity-docker
         - ansible-test-sanity-docker-stable-2.9
-        - ansible-test-sanity-docker-stable-2.10
```
pabelanger, line 123: we still want sanity jobs, just replacing integration testing.

PR #898, opened 2021-05-05
ansible-test: match changef tests

goneri:
When `ansible_test_changed` is enabled, we should also identify
the test modifications.

### DIFF
--- a/roles/ansible-test/tasks/ansible_test_changed.yaml
+++ b/roles/ansible-test/tasks/ansible_test_changed.yaml
@@ -4,7 +4,7 @@
     chdir: "{{ ansible_test_location }}"
     executable: /bin/bash
   shell: |
-    for i in $(git diff origin/main --name-only|egrep '^plugins/.*.py'|xargs -r basename -s .py); do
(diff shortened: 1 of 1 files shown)


roles/ansible-test/tasks/ansible_test_changed.yaml
```diff
@@ -4,7 +4,7 @@
     chdir: "{{ ansible_test_location }}"
     executable: /bin/bash
   shell: |
-    for i in $(git diff origin/main --name-only|egrep '^plugins/.*.py'|xargs -r basename -s .py); do
+    for i in $(git diff origin/main --name-only|egrep '^plugins/.*.py'|xargs -r basename -s .py) $(git diff origin/main --name-only|sed -n 's#tests/integration/targets/\(.*\)/.*#\1#p'); do
```
pabelanger, line 7: drive by review:

Is there a reason to hard code 'origin/main' ?  Rather then using the branch checked out by zuul?

```
{{ zuul.branch }}
```

goneri: Thanks for the suggestion.

PR #910, opened 2021-05-07
Add remaining FRR unit jobs

NilashishC:
Signed-off-by: NilashishC <nilashishchakraborty8@gmail.com>

PR #913, opened 2021-05-07
Remove python38 unit tests

GomathiselviS:

Unit tests in python38 are already covered by network-ee tests.

### DIFF
--- a/zuul.d/cisco-nxos-jobs.yaml
+++ b/zuul.d/cisco-nxos-jobs.yaml
@@ -39,15 +39,6 @@
       ansible_test_integration_targets: "nxos_.*"
 
 - job:
-    name: ansible-test-units-nxos
-    parent: ansible-test-units-base
-    required-projects:
-      - name: github.com/ansible-collections/cisco.nxos
-    timeout: 3600
-    vars:
-      ansible_collections_repo: github.com/ansible-collections/cisco.nxos
-
-- job:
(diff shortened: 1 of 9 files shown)


zuul.d/cisco-nxos-jobs.yaml
```diff
@@ -38,15 +38,6 @@
...
 
-- job:
-    name: ansible-test-units-nxos
-    parent: ansible-test-units-base
-    required-projects:
-      - name: github.com/ansible-collections/cisco.nxos
-    timeout: 3600
-    vars:
-      ansible_collections_repo: github.com/ansible-collections/cisco.nxos
-
```
pabelanger, line 49: We should also be running 2.7, 3.5, 3.6, 3.7 unit testing on nxos. Is that not the case?

zuul.d/project-templates.yaml
```diff
@@ -758,7 +752,6 @@
         - ansible-test-sanity-docker-stable-2.9
         - ansible-test-sanity-docker-stable-2.10
         - ansible-test-sanity-docker-stable-2.11
-        - ansible-test-units-frr
```
pabelanger, line 761: looks like we didn't delete this job... also same comment for nxos, we are missing some unit test coverage here.

GomathiselviS: jobs added in https://github.com/ansible/ansible-zuul-jobs/pull/910

PR #915, opened 2021-05-07
AWS should only be sanity tested on python3

jillr:
We're dropping python2 support in the next collection release, for both collections

### DIFF
--- a/zuul.d/ansible-cloud-jobs.yaml
+++ b/zuul.d/ansible-cloud-jobs.yaml
@@ -323,6 +323,7 @@
       ansible_test_python: 3.6
       ansible_test_changed: true
 
+#### units
 - job:
     name: ansible-test-units-community-aws-python38
     parent: ansible-test-units-base-python38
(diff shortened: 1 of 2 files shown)


jillr:
@pabelanger @GomathiselviS would either of you be able to review?  And also advise - I'm not sure if the way I've gone about this is the best approach for the objective!  (I'd prefer not to have to maintain hundreds of sanity ignores, especially for community.aws which has almost 200 modules)

pabelanger:
> @pabelanger @GomathiselviS would either of you be able to review? And also advise - I'm not sure if the way I've gone about this is the best approach for the objective! (I'd prefer not to have to maintain hundreds of sanity ignores, especially for community.aws which has almost 200 modules)

you only want to test python3+ in the docker container right?

if so, we don't need to create new jobs for that. We can expose a new flag to ansible-test-sanity job to allow the user to pick the version of python to test against.

jillr:
@pabelanger Yep - the AWS collections will be py3.6+.  So I'd edit the base `ansible-test-sanity-docker` (et al) jobs to take `ansible_test_python:` and then put that in zuul.d/ansible-cloud-jobs.yaml, with just that var specified?

pabelanger:
So, i think what we need to do is update our ansible-test command to use

ansible-test sanity --requirements --docker -v --python=<list of versions>

Do you mind confirming locally, that does what you are looking for?

pabelanger:
https://github.com/ansible/ansible-zuul-jobs/blob/master/roles/ansible-test/tasks/init_test_options.yaml#L47-L57

is why it doesn't work today. Which we can fix

jillr:
> So, i think what we need to do is update our ansible-test command to use
> 
> ansible-test sanity --requirements --docker -v --python=3.6,3.7,3.8
> 
> Do you mind confirming locally, that does what you are looking for?

ansible-test doesn't take a list for --python
`ansible-test sanity: error: argument --python: invalid choice: '3.6,3.7,3.8' (choose from '2.6', '2.7', '3.5', '3.6', '3.7', '3.8', '3.9', '3.10', 'default')`

jillr:
@pabelanger Should I submit some change to init_test_options.yaml?  I'm not sure what our next step here is.

zuul.d/ansible-cloud-jobs.yaml
```diff
@@ -335,10 +335,195 @@
...
+    timeout: 3600
+    vars:
+      ansible_test_collections: true
+      ansible_collections_repo: "{{ zuul.project.canonical_name }}"
+      ansible_test_command: sanity
+      ansible_test_docker: true
+      ansible_test_enable_ara: false
+      ansible_test_python: 3.6
+
+- job:
```
goneri, line 366: Since basically, all these jobs are 99% the same, I would inherit them from the first in the list and just overwrite the two different keys.

jillr: tbh I didn't entirely realize I could do that with `parent`!  Let's see if I did that right or not.  :)

jillr:
I had to disabled the `--docker` option because https://github.com/ansible/ansible-zuul-jobs/blob/master/roles/ansible-test/tasks/init_test_options.yaml#L47-L52.  I don't see any immediate issues with doing so (though my preference would be to isolate the tests, I'm disinclined to make changes to the base role).

jillr:
@pabelanger ready_for_review

PR #927, opened 2021-05-14
Add libssh jobs for Cisco ASA collection

justjais:
Add libssh jobs for Cisco ASA collection


justjais:
cc. @pabelanger plz review and gate

PR #935, opened 2021-05-19
add a template for community.kubernetes

goneri:
This way we can extend the `build-ansible-collection` job definition
to declare the dependency on `kubernetes.core`.


goneri:
recheck

PR #941, opened 2021-05-20
DNM:Fix Retry_limit issue

ashwini-mhatre:
fix retry_limit issue


playbooks/ansible-network-iosxr-appliance/files/bootstrap.yaml
```diff
@@ -36,6 +36,14 @@
...
       with_inventory_hostnames: appliance
 
+    - name: Ensure Newly generated remote SSH host keys are known
+      shell: "ssh-keyscan -v -p 22 {{ hostvars[item].nodepool.private_ipv4 }} >> /home/zuul/.ssh/known_hosts"
+      register: r
+      until: r.rc == 0
+      retries: 10
+      delay: 60
+      with_inventory_hostnames: appliance
+
```
pabelanger, line 46: So, if we have to do this, then it is very likely the hypervisor iosxr is running on is too slow.  When we moved from iosxr 6 to 7, did the cpu requirements change?

If this is happening in specific regions, it is very possible we just need to bump the resources in each region.

ashwini-mhatre: Cisco IOS XRv 9000 Router requires 16GB of RAM, 45GB hard disk, and 4 vCPUs. This is specified in following document https://www.cisco.com/c/en/us/td/docs/routers/virtual-routers/configuration/guide/b-xrv9k-cg/b-xrv9k-cg_chapter_011.html#task_1B88FA19EBBE4C80861628DED043256D

ashwini-mhatre: @pabelanger you are might be right. because this issue is coming randomly so. I am suspecting as iosxr is slow its taking long time for booting. I will check with gomathi which flavor she is using.

ashwini-mhatre:
recheck

Issue #934, opened 2021-05-19
Wrong collections path in galaxy-importer job


jillr:
After https://github.com/ansible-network/releases/pull/17 stopped bringing in the entire Ansible package it became apparent that the incorrect collections path is being used in this job.  Collections which use doc fragments from other collections have been failing CI for all jobs since (ie; community.aws used amazon.aws fragments, and community.okd uses fragments from kubernetes.core).

https://bfcb0036379da16f975f-4282179313616d046bba66d871c7a9ef.ssl.cf5.rackcdn.com/579/16b009a77e4f66d679b0308972f29c92577bf4d3/check/ansible-galaxy-importer/bf50899/job-output.html#l779

In irc we determined this has to do with the collections path
https://gist.github.com/jillr/f2a01a25cc902a1a89194a2dccecf749

goneri:
It's a bug in `galaxy-importer`. I just opened the issue: https://issues.redhat.com/browse/AAH-650

IMO, the best way to address the problem for now is to embed the fragments in all the collections.

PR #943, opened 2021-05-20
ignore ansible-galaxy-importer in some cases

goneri:
Ignore `ansible-galaxy-importer` for:

- community.aws
- community.okd

See: https://github.com/ansible/ansible-zuul-jobs/issues/934

### DIFF
--- a/zuul.d/project-templates.yaml
+++ b/zuul.d/project-templates.yaml
@@ -103,6 +103,8 @@
               - name: github.com/ansible-collections/ansible.utils
               - name: github.com/ansible-collections/community.general
         - ansible-test-cloud-integration-aws-py36
+        - ansible-galaxy-importer:
+            voting: false
     gate:
       jobs:
         - ansible-test-sanity-docker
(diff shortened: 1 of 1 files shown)


zuul.d/project-templates.yaml
```diff
@@ -116,6 +118,8 @@
               - name: github.com/ansible-collections/ansible.utils
               - name: github.com/ansible-collections/community.general
         - ansible-test-cloud-integration-aws-py36
+        - ansible-galaxy-importer:
+            voting: false
```
pabelanger, line 122: lets remove non-voting from gate, since results are not used for merging

goneri: I'm not sure I understand. `ansible-galaxy-importer` job also runs during the gating.

pabelanger: non-voting jobs in gate, are a waste of resources. So, rather making a job non-voting, removing it is better. And nice on nodepool.

goneri: The job is pulled by `publish-to-galaxy` and I don't want to change project-config just for this temporary fix. I don't see any easy way to disable a job?

PR #956, opened 2021-06-01
WIP - kubernetes.core molecule test job

abikouo:
add new job to run ``kubernetes.core`` molecule tests when running ci tests for ``cloud.common``


playbooks/ansible-cloud/k8s/pre.yaml
```diff
@@ -0,0 +1,50 @@
...
+    - name: Install collections using symlink
+      ansible.builtin.file:
+        state: link
+        path: '{{ item.dest }}'
+        src: '{{ ansible_user_dir }}/{{ zuul.projects[item.src].src_dir }}'
+      with_items:
+        - src: github.com/ansible-collections/kubernetes.core
+          dest: ~/.ansible/collections/ansible_collections/kubernetes/core
+        - src: github.com/ansible-collections/cloud.common
+          dest: ~/.ansible/collections/ansible_collections/cloud/common
```
pabelanger, line 30: we have a role to do this called deploy-artifacts: https://github.com/ansible/ansible-zuul-jobs/blob/master/roles/deploy-artifacts/tasks/main.yaml

You can use that, along with the build-ansible-collection job. You can look to other network jobs to see how it works.

abikouo: This role is running ``ansible-galaxy collection install (collection)`` which will install collection based on ``galaxy.yml`` or ``Manifest.json`` which is not what I intend, I am trying to use the ``master`` branch of the collection

pabelanger: This is not entirely true, we build all collections ourself in the build-ansible-collection job, which will pull content from git rather then galaxy. We then, install the collection via a local tarball.

Basically, it will give you the same you are doing here, expect using a standard way of installing content.

abikouo: ok I will then use this, thanks

roles/setup-kind/tasks/main.yml
```diff
@@ -0,0 +1,24 @@
+---
+- name: Install Kind
+  ansible.builtin.get_url:
+    url: '{{ "https://kind.sigs.k8s.io/dl/" + kind_release + "/kind-linux-amd64" }}'
+    dest: '{{ kind_path }}'
+  become: true
+
+- name: Install Kubeclt
```
Akasurde, line 8:
```suggestion
- name: Install Kubectl
```

playbooks/ansible-cloud/k8s/run.yaml
```diff
@@ -0,0 +1,9 @@
+---
+- hosts: all
+  vars:
+    ansible_zuul_src_dir: "{{ ansible_user_dir }}/{{ zuul.projects['github.com/ansible/ansible-zuul-jobs'].src_dir }}"
+  tasks:
```
goneri, line 5: Can we use tox here? This was we would be able to delegate the venv creation and the molecule execution to tox and keep the configuration (tox.ini and test-requirements.txt) within the repositor(y|ies) that we test.

Akasurde:
@abikouo Could you please take a look at https://github.com/ansible/ansible-zuul-jobs/pull/998 and modify this PR? Thanks.

abikouo:
merged into #998

PR #957, opened 2021-06-07
don't update vcsa

goneri:
Follow up to a recent Photon Linux update, the REST interface is broken.
We disable the update until we've got a better fix.

See: https://github.com/ansible-collections/community.vmware/issues/875

### DIFF
--- a/playbooks/ansible-cloud/vcenter-appliance/pre.yaml
+++ b/playbooks/ansible-cloud/vcenter-appliance/pre.yaml
@@ -38,9 +38,6 @@
       become: true
     - import_role:
         name: vmware-ci-memory-configuration
-    - name: Update the VCSA packages
-      command: tdnf update -y
-      become: true
 
 - hosts: all
   gather_facts: false


mariolenz:
@goneri I don't understand what you removed here:

>    - name: Update the VCSA packages
>      command: tdnf update -y
>      become: true

Is this run _on the VCSA_? If the answer is "yes", I don't think it's supported to update the VCSA this way. In other words: VMware might release new packages for Photon, but it's quite likely they're not tested for compatibility with vCenter.

So _if_ this is run on the VCSA, you should consider not activating this task again to avoid future trouble.

goneri:
> @goneri I don't understand what you removed here:
> 
> > * name: Update the VCSA packages
> >   command: tdnf update -y
> >   become: true
> 
> Is this run _on the VCSA_? If the answer is "yes", I don't think it's supported to update the VCSA this way. In other words: VMware might release new packages for Photon, but it's quite likely they're not tested for compatibility with vCenter.
> 
> So _if_ this is run on the VCSA, you should consider not activating this task again to avoid future trouble.

Yes, probably. The thing is, until recently, this was enough to fix one of the appliance/networking REST end-points :-).

mariolenz:
> The thing is, until recently, this was enough to fix one of the appliance/networking REST end-points :-).

Well, let's say that sometimes a quick'n'dirty fix turns out to be _really_ dirty :-P

PR #962, opened 2021-06-18
Split AWS jobs even more

jillr:
Group 2 is timing out a lot

### DIFF
--- a/zuul.d/ansible-cloud-jobs.yaml
+++ b/zuul.d/ansible-cloud-jobs.yaml
@@ -323,31 +323,37 @@
       ansible_test_enable_ara: false
       ansible_test_python: 3.6
       ansible_test_changed: true
-      ansible_test_split_in: 4
+      ansible_test_split_in: 5
 
 - job:
-    name: ansible-test-cloud-integration-aws-py36_1_of_4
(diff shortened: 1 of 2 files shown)


jillr:
@goneri  Can you confirm there's no other way I can do this?  On community.aws right now with 5 splits I have 3 jobs that take <23 minutes but 1 that takes more than an hour - so I have to try 6 splits.  We're going to be at risk of hitting API call thresholds at this rate. 

I know which test suites shouldn't be in the same group but I don't know any way to manually assign tests to groups in zuul like we could with shippable.  Is there something I could be doing to have more control over this?

goneri:
> @goneri Can you confirm there's no other way I can do this? On community.aws right now with 5 splits I have 3 jobs that take <23 minutes but 1 that takes more than an hour - so I have to try 6 splits. We're going to be at risk of hitting API call thresholds at this rate.
> 
> I know which test suites shouldn't be in the same group but I don't know any way to manually assign tests to groups in zuul like we could with shippable. Is there something I could be doing to have more control over this?

We can certainly improve this. The split is done here: https://github.com/ansible/ansible-zuul-jobs/blob/master/roles/ansible-test/tasks/split_targets.yaml#L2

PR #968, opened 2021-06-26
fix: even distribution for test targets

oukooveu:
first `ansible_test_split_in` jobs will be skipped due to current logic


oukooveu:
closed because of https://github.com/ansible/ansible-zuul-jobs/pull/973

oukooveu:
closed because of #973

goneri:
Oh @oukooveu! I sorry I missed this patch.

oukooveu:
> Oh @oukooveu! I sorry I missed this patch.

@goneri no problem, your patch is better but we need to make it working

PR #969, opened 2021-06-28
Sort the list of modules based on a hash of their name

tremble:
The AWS modules have a lot of 'slow' tests with related names.  If we hash the list we reduce the chance that they clump together


goneri:
Nice. I'm curious to see if they won't break some tests.

goneri (approved):

PR #973, opened 2021-06-28
ansible-test/split_targets: improve the start/end computation

goneri:
Simplify the logic and correctly compute the end of the first job
in case the number of targets is lower than `ansible_test_split_in`.

### DIFF
--- a/roles/ansible-test/tasks/split_targets.yaml
+++ b/roles/ansible-test/tasks/split_targets.yaml
@@ -10,8 +10,11 @@
 - set_fact:
     number_entries: "{{ ansible_test_targets.stdout_lines|length }}"
 - set_fact:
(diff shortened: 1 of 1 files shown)


goneri:
recheck

goneri:
recheck

oukooveu:
are you sure this will work as expected? for example for `ansible_test_split_in = 3` and `number_entries = 10` this produces ranges: 0-4, 4-8 and 8-12. 

i can suggest following change:
```
_end_at: "{{ _start_at|int + _iter_by|int - 1 if ansible_test_do_number|int == _iter_by else number_entries - 1 }}"
```
and this produces 0-3, 4-7 and 8-9 for the same input.

PR #979, opened 2021-07-01
Remove hashing of target names

tremble:
The *prefix* of the name is what tends to be similar for the slow tests, now that we pull every x-th entry we'll get a better distribution without hashing

### DIFF
--- a/roles/ansible-test/tasks/split_targets.yaml
+++ b/roles/ansible-test/tasks/split_targets.yaml
@@ -11,16 +11,11 @@
   failed_when: ansible_test_targets.rc not in [0, 1]
 - set_fact:
     number_entries: "{{ ansible_test_targets.stdout_lines|length }}"
-# Slow tests tend to be closely related.  Hash the names to produce a repeatable
-# list that avoids bringing those tests all into one group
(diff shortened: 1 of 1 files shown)


roles/ansible-test/tasks/split_targets.yaml
```diff
@@ -11,16 +11,11 @@
...
 - set_fact:
     number_entries: "{{ ansible_test_targets.stdout_lines|length }}"
-# Slow tests tend to be closely related.  Hash the names to produce a repeatable
-# list that avoids bringing those tests all into one group
 - set_fact:
-    _hashed_targets: "{{ _hashed_targets + [ { 'name': item, 'hash': (item | hash('md5')) } ] }}"
-  loop: '{{ ansible_test_targets.stdout_lines }}'
-- set_fact:
-    sorted_test_targets: "{{ _hashed_targets | sort(attribute='hash') | map(attribute='name') | list }}"
+    sorted_test_targets: "{{  ansible_test_targets.stdout_lines | sort() | list }}"
```
goneri, line 15: We don't need `sorted_test_targets` anymore.

PR #981, opened 2021-07-06
Remove integration tests jobs from check and gate pipeline.

ashwini-mhatre:
* Remove integration tests jobs from check and gate pipeline until retry_limit issue is fixed.
* instead of making this jobs nonvoting removed this jobs until we get valid fix for retry_limit issue to save resources.


ashwini-mhatre:
We have removed iosxr integration jobs beacuse of https://storyboard.openstack.org/#!/story/2009226

PR #983, opened 2021-07-08
bindep.txt: install python3-libselinux on Fedora

goneri:


mariolenz:
@goneri Do you think we still need this? It looks like ansible-collections/community.vmware#949 already fixed ansible-collections/community.vmware#939.

goneri:
Cosing as per @mariolenz's comment

PR #991, opened 2021-07-16
Add test command specific pip requirements/constraints

tremble:
Ansible-test normally pulls in test-command specific python requirements.  This means that we can do things like forcing an old version of boto3/botocore during tests to ensure unsupported API calls are tested for.

Depends-On: https://github.com/ansible-collections/amazon.aws/pull/404
Depends-On: https://github.com/ansible-collections/community.aws/pull/644

### DIFF
--- a/roles/ansible-test/tasks/init_collection.yaml
+++ b/roles/ansible-test/tasks/init_collection.yaml
@@ -4,8 +4,36 @@
     _test_location: "{{ ansible_test_collection_dir }}/{{ ansible_test_collection_namespace }}/{{ ansible_test_collection_name }}"
     _test_cfg_location: "{{ ansible_test_collection_dir }}/{{ ansible_test_collection_namespace }}/{{ ansible_test_collection_name }}/tests/integration/{{ ansible_test_test_command }}.cfg"
 
+- name: Setup minimum test requirements
+  set_fact:
+    _test_requirements: '-r {{ _test_location }}/requirements.txt -r {{ _test_location }}/test-requirements.txt'
+
+- name: Check for existence of test command specific requirements/constraints
+  when: ansible_test_test_command in _file_locations
+  vars:
+    _file_locations:
+      integration: '{{ _test_location }}/tests/integration'
+      units: '{{ _test_location }}/tests/unit'
+  block:
+    - name: Check for test command constraints file
+      stat:
+        path: '{{ _file_locations[ansible_test_test_command] }}/constraints.txt'
+      register: _extra_constraints_stat
+    - name: Check for test command requirements file
+      stat:
+        path: '{{ _file_locations[ansible_test_test_command] }}/requirements.txt'
+      register: _extra_requirements_stat
+    - name: Add extra constraints.txt
+      when: _extra_constraints_stat.stat.exists
+      set_fact:
+        _test_requirements: '{{_test_requirements}} -c {{ _file_locations[ansible_test_test_command] }}/constraints.txt'
+    - name: Add extra requirements.txt
+      when: _extra_requirements_stat.stat.exists
+      set_fact:
+        _test_requirements: '{{_test_requirements}} -r {{ _file_locations[ansible_test_test_command] }}/requirements.txt'
+
 - name: Install python requirements
-  shell: "{{ ansible_test_venv_path }}/bin/pip install -r {{ _test_location }}/requirements.txt -r {{ _test_location }}/test-requirements.txt"
+  shell: "{{ ansible_test_venv_path }}/bin/pip install {{ _test_requirements }}"
 
 - name: Copy potential cloud provider configuration for ansible-test in the collection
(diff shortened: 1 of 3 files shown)


pabelanger:
Thanks for this patch, in fact I am hoping next week to raise this discuss on creating global constraints file for all projects to opt into. Allowing for both upper / lower constraints for testing.

tremble:
@pabelanger Was there a decision on the direction you wanted to take here?

jillr:
Chatted with pabelanger; we need to test the cisco.ios collection against this patch to ensure nothing will break in the network collections before we merge this. Global constraints are still desirable but the work to do it is not currently planned/prioritized.

tremble:
@jillr @pabelanger 

Tested : https://github.com/ansible-collections/cisco.ios/pull/418

Only change that cisco.ios would see is that the unit tests would install some extra deps.  This change only has an effect if the requirements.txt or constraints.txt files exist (as used to be the case pre-zuul migration when everything used the ansible-test container-based tests)

roles/ansible-test/tasks/init_collection.yaml
```diff
@@ -4,8 +4,36 @@
...
+- name: Setup minimum test requirements
+  set_fact:
+    _test_requirements: '-r {{ _test_location }}/requirements.txt -r {{ _test_location }}/test-requirements.txt'
+
+- name: Check for existence of test command specific requirements/constraints
+  when: ansible_test_test_command in _file_locations
+  vars:
+    _file_locations:
+      integration: '{{ _test_location }}/tests/integration'
+      units: '{{ _test_location }}/tests/unit'
```
goneri, line 16: Can we standardize on one of these two files?

tremble: We can, I'd gone for splitting it like this to be consistent with the behaviour of the containerized ansible test

jillr: Not really as long as we use ansible-test to execute the tests.  These locations are picked up by ansible-test in and used to install the needed requirements for the given type of test.

pabelanger: Ideally, the way I see this working is we create github.com/ansible-collections/requirements repo. In it, we store an upper-constraints.txt file, along with other lower constraints and caps.

unit / sanity / integration tests should use the same constraints, along with all the collections we ship in an execution environment.  The goal, is to not only ensure each collection works with the specific libraries but all of our collections are installable together with the same constraints.

https://docs.openstack.org/project-team-guide/dependency-management.html really explains what I am hoping we get too.

tremble: The use case I was aiming for with the test constraints was to ensure that the AWS modules work with the minimum versions of boto3/botocore we claim to support.  Amazon add new features very rapidly, and when we don't explicitly check it's easy to accidentally rely on a newer feature and break things for people trying to use the minimum version we say we support.

As such the test constraint files are going to be specific to the version of the collection, this doesn't work particularly well for storing the test constraints outside of the collection repo.

roles/ansible-test/tasks/init_collection.yaml
```diff
@@ -4,8 +4,36 @@
     _test_location: "{{ ansible_test_collection_dir }}/{{ ansible_test_collection_namespace }}/{{ ansible_test_collection_name }}"
     _test_cfg_location: "{{ ansible_test_collection_dir }}/{{ ansible_test_collection_namespace }}/{{ ansible_test_collection_name }}/tests/integration/{{ ansible_test_test_command }}.cfg"
 
+- name: Setup minimum test requirements
+  set_fact:
+    _test_requirements: '-r {{ _test_location }}/requirements.txt -r {{ _test_location }}/test-requirements.txt'
```
goneri, line 9: Can we set a new `ansible_test_test_requirements_parameters` in ansible-test/defaults/main.yaml  with`'-r {{ _test_location }}/requirements.txt -r {{ _test_location }}/test-requirements.txt'` as the default value. This way we can overload the value for each project and we avoid this block? And this would remove the need of this block.

goneri (approved):

goneri:
Job is failing with this PR:
https://github.com/ansible-collections/amazon.aws/pull/505
https://dashboard.zuul.ansible.com/t/ansible/build/25bf6109237e4a7b95852b61de133159

tremble:
@goneri yeah, amazon.aws # 404 lands the constraints file

goneri:
The Depends-On is in the wrong direction since we should land https://github.com/ansible-collections/amazon.aws/pull/404 first.

> @goneri yeah, amazon.aws # 404 lands the constraints file

The Depends-On: were going in the wrong direction.

tremble:
> The Depends-On: were going in the wrong direction.

That was the easiest way to test everything, fair point that they need merging before this one.

tremble:
@goneri @alinabuzachis :
Deps are now in place.  Any chance we can get this merged?

PR #997, opened 2021-07-20
DNM - vmware - rollback recent nodeset changes

goneri:


mariolenz:
I think this PR should be closed because it's probably outdated.

PR #1002, opened 2021-07-21
kubernetes.core: add molecule jobs

Akasurde:


abikouo (changes requested):
would be better to rename the file ``playbooks/ansible-cloud/k8s/run.yaml`` to ``playbooks/ansible-cloud/k8s/integration.yaml`` in order to have one file per operation like the molecule one ``playbooks/ansible-cloud/k8s/molecule.yaml``

PR #1004, opened 2021-07-21
Make eos jobs as non voting

GomathiselviS:

With the upgrade of eos image in CI , we are fixing the failures seen due to changes in cli syntax. Until the fixes are merged, the eos jobs will remain non-voting.

### DIFF
--- a/zuul.d/project-templates.yaml
+++ b/zuul.d/project-templates.yaml
@@ -191,37 +191,37 @@
     name: ansible-collections-arista-eos
     check:
       jobs: &ansible-collections-arista-eos-jobs
-        - ansible-test-network-integration-eos-httpapi-python27-stable211
+        - ansible-test-network-integration-eos-httpapi-python27-stable211:
             voting: false 
-        - ansible-test-network-integration-eos-httpapi-python27-stable29
+        - ansible-test-network-integration-eos-httpapi-python27-stable29:
             voting: false
-        - ansible-test-network-integration-eos-network_cli-python27-stable211-scenario01
+        - ansible-test-network-integration-eos-network_cli-python27-stable211-scenario01:
             voting: false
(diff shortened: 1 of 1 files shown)


pabelanger:
you really should revert your change to migrate to latest EOS images, until your testing is green.  Otherwise, we now have a gap in testing

sc68cal:
I would suggest upgrading the CI job that votes up to an EOS version that supports both the old CLI syntax as well as the new syntax, and then have a non-voting CI job for EOS 4.24

I _think_ EOS version 4.22 is still supported, and has support for the new syntax while still honoring the old syntax.

That would be my suggestion

pabelanger:
I think there is a different way to handle this, which I can hopefully summarize. 

But moving all these jobs to non-voting, any PR that is run will not report back if the results are broken or not.  This seems like the wrong approach.  What can be done, is either:

  First get all CI testing green, then bring the new images online (4.24.0) via depends-on.  We are in the position to not actually first merge the code for testing, given our pre-merge testing ability.
  
  Second, disable the specific tests that will break, inside the collection, bring online new images, then work to re-enable testing.  I am not a fan of this approach, because we basically stop all testing, cut over, then work to bring it back online.  This to me is too aggressive, especially since we have option 1 (which likely will be more PRs).
  
The other issue, which I should comment in the other PR. Is I don't think we should remove 4.22.0 images when we add 4.24.0 images.  What we should likely do, is keep 4.22.0 images on stable branches, until that version of ansible is EOL, and add the latest 4.24.0 image to the next ansible-core stable branch release (2.13?).

Otherwise, we drop support for an appliance mid release, which impacts downstream supportability to users.

GomathiselviS:
Closing this PR, because of the above-said reasons.

PR #1010, opened 2021-07-26
k8s job on cloud.common

abikouo:
Testing ``kubernetes.core`` while validating ``cloud.common``

### DIFF
--- a/playbooks/ansible-cloud/k8s/pre.yaml
+++ b/playbooks/ansible-cloud/k8s/pre.yaml
@@ -15,13 +15,13 @@
         _virtualenv_options: "--python python{{ ansible_test_python }}"
 
     - name: Create virtualenv for ansible-test
-      shell: "virtualenv {{ _virtualenv_options }} ~/venv"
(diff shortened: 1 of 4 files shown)


playbooks/ansible-cloud/k8s/pre.yaml
```diff
@@ -15,13 +15,13 @@
         _virtualenv_options: "--python python{{ ansible_test_python }}"
 
     - name: Create virtualenv for ansible-test
-      shell: "virtualenv {{ _virtualenv_options }} ~/venv"
+      shell: "virtualenv {{ _virtualenv_options }} {{ ansible_test_venv_path }}"
 
     - name: Install selinux into virtualenv
-      shell: ~/venv/bin/pip install selinux "setuptools<50.0.0"
+      shell: '{{ ansible_test_venv_path }}/bin/pip install selinux "setuptools<50.0.0"'
```
Akasurde, line 21: We are installing SELinux Python Bindings at line 27 so this is not required.

PR #1011, opened 2021-07-27
kubernetes.core lint job

abikouo:
add job to validate kubernetes.core lint using tox


abikouo:
@Akasurde @goneri

playbooks/ansible-cloud/k8s/tox.yaml
```diff
@@ -0,0 +1,23 @@
+---
+- hosts: all
+  tasks:
+    - name: Setup ensure-tox role
+      include_role:
+        name: ensure-tox
+
+    - name: Install SELinux Python bindings using yum
```
goneri, line 8: This binary depends on a specific version on Python and are likely to create trouble if we change this version later.

Most of the time we don't use any SELinux specific feature and the tests run well without it. In this case an alternative is to just mock selinux.
https://pypi.org/project/selinux-please-lie-to-me/

zuul.d/project-templates.yaml
```diff
@@ -720,6 +720,7 @@
         - ansible-test-units-kubernetes-core-python37
         - ansible-test-integration-kubernetes-core
         - ansible-test-molecule-kubernetes-core
+        - ansible-test-lint-kubernetes-core
```
goneri, line 723: You can just use the existing `ansible-tox-linters` template.

abikouo: I did not know that it was existing, thanks

playbooks/ansible-cloud/k8s/tox.yaml
Akasurde: Without SElinux binding, I saw failures.

goneri: It should be find if you replace selinux with this package. BTW you can now remove  playbooks/ansible-cloud/k8s/tox.yaml from the PR :-).

goneri (changes requested):

goneri:
Thanks @abikouo :-)

goneri (approved):

PR #1012, opened 2021-07-27
[Kubernetes.core] run molecule tests on multiple ansible version

abikouo:
add molecule jobs for ansible stable-2.9, stable-2.10 and stable-2.11

### DIFF
--- a/zuul.d/ansible-cloud-jobs.yaml
+++ b/zuul.d/ansible-cloud-jobs.yaml
@@ -449,39 +449,26 @@
       ansible_collections_repo: github.com/ansible-collections/kubernetes.core
 
 - job:
-    name: ansible-test-integration-kubernetes-core
-    dependencies:
-      - name: build-ansible-collection
-    pre-run:
-      - playbooks/ansible-cloud/k8s/pre.yaml
-    run: playbooks/ansible-cloud/k8s/integration.yaml
-    required-projects:
-      - name: github.com/ansible/ansible
-      - name: github.com/ansible-collections/kubernetes.core
-    timeout: 5400
-    nodeset: centos-7-4vcpu
-    vars:
-      ansible_collections_repo: github.com/ansible-collections/kubernetes.core
-      ansible_test_python: 3.6
-      ansible_test_command: integration
-      ansible_test_collections: true
-      ansible_test_venv_path: "~/venv"
-
-- job:
     name: ansible-test-molecule-kubernetes-core
     dependencies:
       - name: build-ansible-collection
     pre-run:
+      - playbooks/ansible-network-appliance-base/pre.yaml
       - playbooks/ansible-cloud/k8s/pre.yaml
-    run: playbooks/ansible-cloud/k8s/molecule.yaml
+    run: playbooks/ansible-cloud/k8s/run.yaml
     required-projects:
       - name: github.com/ansible/ansible
-      - name: github.com/ansible-collections/kubernetes.core
     timeout: 5400
-    nodeset: centos-7-4vcpu
+    nodeset: kind-centos
+    # NOTE: we set ansible_network_os with some generic
+    # value to pass the ansible-network playbooks
+    host-vars:
+      fedora-34:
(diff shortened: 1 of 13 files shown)


abikouo:
recheck

zuul.d/ansible-cloud-jobs.yaml
```diff
@@ -473,6 +473,67 @@
...
+    nodeset: centos-7-4vcpu
+    vars:
+      ansible_collections_repo: github.com/ansible-collections/kubernetes.core
+      ansible_test_python: 3.6
+      ansible_test_command: integration
+      ansible_test_collections: true
+      ansible_test_venv_path: "~/venv"
+
+- job:
+    name: ansible-test-molecule-kubernetes-core-stable-2.10
```
goneri, line 508: I think we can just skip 2.10. It's not an LTS and it's superseded by 2.11 now.

zuul.d/ansible-cloud-jobs.yaml
```diff
@@ -473,6 +473,67 @@
...
+    nodeset: centos-7-4vcpu
+    vars:
+      ansible_collections_repo: github.com/ansible-collections/kubernetes.core
+      ansible_test_python: 3.6
+      ansible_test_command: integration
+      ansible_test_collections: true
+      ansible_test_venv_path: "~/venv"
+
+- job:
+    name: ansible-test-molecule-kubernetes-core-stable-2.11
```
goneri, line 528: Both jobs are pretty much the same you can use the `parent` key to inherit from the first one and reduce the information duplication. This is just a suggestion, sometime this also reduces the readability.

abikouo: done
Thanks

zuul.d/ansible-cloud-jobs.yaml
abikouo: ok will do it.

zuul.d/ansible-cloud-jobs.yaml
```diff
@@ -473,12 +471,25 @@
     run: playbooks/ansible-cloud/k8s/molecule.yaml
     required-projects:
       - name: github.com/ansible/ansible
+        override-checkout: devel
```
goneri, line 474: `devel` is already the default.

zuul.d/project-templates.yaml
```diff
@@ -711,18 +713,59 @@
...
+        - ansible-test-molecule-kubernetes-core-210_0
+        - ansible-test-molecule-kubernetes-core-210_1
+        - ansible-test-molecule-kubernetes-core-210_2
+        - ansible-test-molecule-kubernetes-core-210_3
+        - ansible-test-molecule-kubernetes-core-210_4
+        - ansible-test-molecule-kubernetes-core-devel_0
+        - ansible-test-molecule-kubernetes-core-devel_1
+        - ansible-test-molecule-kubernetes-core-devel_2
+        - ansible-test-molecule-kubernetes-core-devel_3
+        - ansible-test-molecule-kubernetes-core-devel_4
```
pabelanger, line 740: What is the idea of doing this?  This is pretty aggressive from a compute / testing POV, given each of these nodes is 4vcpu.  Are you hitting timeout issues?

abikouo: the 4vcpu configuration was originally with the full scenario execution, I will try that with less cpu, knowing that I will need at least 2 as this is running a k8s cluster.

pabelanger: The better way here is to create a new nodeset, with 2 nodes. One for controller, the other appliance. Then you can boot a 1vcpu controller, and 2vcpu appliance (k8s).

We do this pattern of testing for network appliances, and it works well if you need to scale up.

That way, we can drop this complex logic, which is really only needed for limiting testing on aws

abikouo: yes this can be a solution, but we also trying here to speed up the execution, now the full molecule default scenario is running in more than 45 minutes
I will try this

PR #1013, opened 2021-07-28
Make the ansible-test-splitter job only run changed targets

jillr:
We're getting timeouts on some jobs again.  --diff doesn't seem to actually be doing anything, all jobs are running on small changes, but that will take more time to figure out and this is blocking everything.

### DIFF
--- a/playbooks/ansible-test-splitter/run.yaml
+++ b/playbooks/ansible-test-splitter/run.yaml
@@ -1,10 +1,6 @@
 ---
 - hosts: all
   tasks:
-    - name: You are here
-      debug:
-        msg: '{{ ansible_test_splitter__test_changed }}'
-
     - name: Run ansible-test-splitter
       include_role:
         name: ansible-test-splitter
--- a/playbooks/ansible-test-splitter/run.yaml
+++ b/playbooks/ansible-test-splitter/run.yaml
@@ -1,8 +1,12 @@
 ---
 - hosts: all
   tasks:
+    - name: You are here
+      debug:
+        msg: '{{ ansible_test_splitter__test_changed }}'
+
     - name: Run ansible-test-splitter
-      import_role:
+      include_role:
         name: ansible-test-splitter
       vars:
         ansible_test_test_command: "{{ ansible_test_command }}"
--- a/roles/ansible-test-splitter/files/split_targets.py
+++ b/roles/ansible-test-splitter/files/split_targets.py
@@ -3,15 +3,16 @@
 from pathlib import PosixPath
 import sys
 import json
+import subprocess
 
 job_prefix = sys.argv[1]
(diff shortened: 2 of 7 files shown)


jillr:
@abikouo I've cherry picked your change from 1029 into this branch and rebased on master.  Assuming amazon.aws 439 passes (it will likely run past my end of day) does this PR have everything we need to merge from your perspective?  I'm CI vanguard now so you don't need to worry about this one anymore.  :)

abikouo:
> @abikouo I've cherry picked your change from 1029 into this branch and rebased on master. Assuming amazon.aws 439 passes (it will likely run past my end of day) does this PR have everything we need to merge from your perspective? I'm CI vanguard now so you don't need to worry about this one anymore. :)
Thanks @jillr 
I will cancel the 1029. This one can be merged as it fixed the ``test changed-only`` feature even if it is not activated.
By the way enjoy your CI vanguard Role :)

jillr:
Squashed commits to remove all the debug commit messages.  This PR now:
Fixes ansible-test-splitter to only run changed targets
Enables only testing changed for check/gate on AWS collections
Fixes some typos
Pulls in abikuou's minor changes from 1029

Tested in https://github.com/ansible-collections/amazon.aws/pull/439

jillr:
@tremble FYI, when this lands we'll be only tested changed targets.  I'll work on adding an optional check_complete type pipeline next week (so we can call it on-demand like we used to do ci_complete in shippable)

playbooks/ansible-test-splitter/run.yaml
```diff
@@ -1,8 +1,12 @@
 ---
 - hosts: all
   tasks:
+    - name: You are here
```
goneri, line 4: Would you like to keep this in the merged version?

roles/ansible-test-splitter/files/split_targets.py
```diff
@@ -3,21 +3,25 @@
...
 import json
+import argparse
+import os
 
-job_prefix = sys.argv[1]
-if len(sys.argv) == 3:
-    targets_from_cli = sys.argv[2].split(" ")
-else:
-    targets_from_cli = []
+parser = argparse.ArgumentParser()
```
goneri, line 9: Thank you, this is a nice improvement.

playbooks/ansible-test-splitter/run.yaml
jillr: Hehe, no thank you!

roles/ansible-test-splitter/files/split_targets.py
jillr: That was Aubin's commit, I agree it's good work!

goneri (approved):

PR #1018, opened 2021-07-30
Evenly distributed targets to remaining jobs

gravesm:
This will take all the remaining jobs (after the slow ones have been assigned) and evenly distribute them across the remaining jobs. I'm assuming a hard coded number of 10 available jobs, but I don't know how correct that assumption is. The intent here is to avoid the timeouts we've been seeing with the AWS collections.

### DIFF
--- a/roles/ansible-test-splitter/files/split_targets.py
+++ b/roles/ansible-test-splitter/files/split_targets.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python3
 
 from pathlib import PosixPath
-import random
 import sys
 import json
 
@@ -35,8 +34,6 @@ for target in targets.glob("*"):
         batches.append([target.name])
     else:
         regular_targets.append(target.name)
-
-random.shuffle(regular_targets)
 
 slow_jobs = len(batches)
 remaining_jobs = total_jobs - slow_jobs
--- a/roles/ansible-test-splitter/files/split_targets.py
+++ b/roles/ansible-test-splitter/files/split_targets.py
@@ -11,7 +11,7 @@ if len(sys.argv) == 3:
 else:
     targets_from_cli = []
 jobs = [f"{job_prefix}{i}" for i in range(10)]
-targets_per_job = 20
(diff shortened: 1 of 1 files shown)


roles/ansible-test-splitter/files/split_targets.py
```diff
@@ -38,13 +38,13 @@
 
 random.shuffle(regular_targets)
```
tremble, line 39: Given the use of `[x::remaining_jobs]` I'd recommend removing the shuffle here.  We already know that in general slower tests tend to use similar prefixes

jillr (approved):

pabelanger:
@gravesm @tremble @jillr: how slow is slow?  There is a downside to this approach, we boot more test nodes in nodepool. And given rate limit on concurrent jobs that aws has at one time (6) I believe it is actually slowing down testing over the large scale.

Before this merged, how long would a test job run?

pabelanger:
Basically, we should establish a min amount of tests that a job runs say 5. Then if we overflow that, split into the next job. 

Right now, we are booting nodes, just to run a single test... which is fine if the system wasn't busy but when we are at capacity it actually slows things down.  It takes 5-10mins to boot / prepare the node, which is a lot if a test only take 5 mins to run.

tremble:
"slow" tests take at least 10 minutes, some take up to 30 minutes.  The rest generally take between 30 seconds and 5 minutes.  I believe @goneri has written a script to grab the times from the last runs.  Both @jillr and @goneri are out on PTO at the minute.  @jillr's back on Monday and should be able to provide better details (especially after I just pruned a lot of pip virtual environment building)

Back on Shippable we were running these in 5 or 6 groups (manually balanced) each of which was taking between 30 and 60 minutes to run.

Running on 5 nodes automatically balanced we were regularly hitting the 60 minute limit (after which the credentials expire).

@abikouo was working on #1041 which would significantly drop the load by only running tests during the "check" phase if the relevant module changed.

PR #1024, opened 2021-08-05
kubernetes.core job - molecule testing

abikouo:
kubernetes.core: use new node `centos-8-stream` as the process was failing in the previous one

amazon.aws: `check` should test only changed, all others targets will be tests on `gate`

### DIFF
--- a/playbooks/ansible-cloud/k8s/pre.yaml
+++ b/playbooks/ansible-cloud/k8s/pre.yaml
@@ -10,6 +10,14 @@
       include_role:
         name: ensure-docker
 
+    - name: Run ensure-virtualenv role
+      include_role:
+        name: ensure-virtualenv
+
+    - name: Run ensure-pip role
+      include_role:
+        name: ensure-pip
+
     - name: Setup base virtualenv_options
       set_fact:
         _virtualenv_options: "--python python{{ ansible_test_python }}"
--- a/zuul.d/ansible-cloud-jobs.yaml
+++ b/zuul.d/ansible-cloud-jobs.yaml
@@ -405,11 +405,10 @@
       - name: github.com/ansible/ansible
       - name: github.com/ansible-collections/kubernetes.core
     timeout: 5400
-    nodeset: centos-7-4vcpu
+    nodeset: centos-8-stream
     vars:
       ansible_collections_repo: github.com/ansible-collections/kubernetes.core
       ansible_test_python: 3.6
-      ansible_test_command: integration
       ansible_test_collections: true
       ansible_test_venv_path: "~/venv"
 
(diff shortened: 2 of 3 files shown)


zuul.d/nodesets.yaml
```diff
@@ -29,6 +29,12 @@
       - name: fedora-34
         label: fedora-34-1vcpu
 
+- nodeset:
+    name: fedora-latest-4vcpu
+    nodes:
+      - name: fedora-34
+        label: fedora-34-4vcpu
+
```
pabelanger, line 37: what was failing on centos-8? I am not sure we want to move all testing to fedora, given it has a support window of 12 months.

I added centos-8-stream images on the weekend, does that nodeset work for you?

abikouo: we are planning to use tox for the kubernetes.core collection, but the centos-7 version of molecule is not up to date, I think we can upgrade it but I don't know the impact on others jobs running with.

abikouo: centos-8 has only 1 vcpu molecule run will timeout

pabelanger: can you try centos-8-stream-1vcpu

I'll create 2vcpu now, so we can also test it.  I really want to get off fedora for testing as it is too fast moving.

pabelanger:
@abikouo https://github.com/ansible-network/windmill-config/pull/875 adds new 2vcpu for centos-8-stream. Lets give that a test please once merged.

zuul.d/nodesets.yaml
```diff
@@ -23,6 +23,12 @@
       - name: centos-8-stream
         label: centos-8-stream
 
+- nodeset:
```
pabelanger, line 26: lets remove this nodeset, we do not need to create a new one.

zuul.d/ansible-cloud-jobs.yaml
```diff
@@ -424,10 +423,9 @@
       - name: github.com/ansible/ansible
       - name: github.com/ansible-collections/kubernetes.core
     timeout: 5400
-    nodeset: centos-7-4vcpu
+    nodeset: centos-8-stream-2vcpu
```
pabelanger, line 426: lets change this to centos-8-stream

playbooks/ansible-cloud/k8s/pre.yaml
```diff
@@ -10,6 +10,11 @@
       include_role:
         name: ensure-docker
 
+    - name: Install virtualenv
+      pip:
+        name: virtualenv
+      become: true
+
```
pabelanger, line 17: you can change this to ensure-virtualenv role, like we do with ensure-docker above

playbooks/ansible-cloud/k8s/pre.yaml
```diff
@@ -22,7 +27,8 @@
 
     - name: Install SELinux Python bindings using yum
       yum:
-        name: libselinux-python3
+        name:
+          - libselinux-python3
```
pabelanger, line 31: please revert, no need to modify

PR #1028, opened 2021-08-07
When running AWS sanity tests limit which python versions we test against

tremble:
AWS collections don't support Python<3.6, explicitly limit which versions of python we run our compilation tests against.

### DIFF
--- a/roles/ansible-test/tasks/init_test_options.yaml
+++ b/roles/ansible-test/tasks/init_test_options.yaml
@@ -49,7 +49,6 @@
     ansible_test_options: "{{ ansible_test_options }} --python {{ ansible_test_python }}"
   when:
     - ansible_test_python
-    - not ansible_test_docker
 
 - name: Setup --docker option
   set_fact:
--- a/zuul.d/ansible-cloud-jobs.yaml
+++ b/zuul.d/ansible-cloud-jobs.yaml
@@ -389,6 +389,9 @@
       ansible_test_command: sanity
       ansible_test_enable_ara: false
       ansible_test_python: 3.6
+      ansible_test_sanity_skiptests:
+        - 'future-import-boilerplate'
+        - 'metaclass-boilerplate'
 
 - job:
     name: ansible-test-sanity-aws-ansible-2.9-python36


pabelanger:
Not sure I understand, you don't want to run python3.6 sanity tests, is that right?

tremble:
@pabelanger if you take a look at https://dashboard.zuul.ansible.com/t/ansible/build/8da0a886f0a24f20b7045d072883a935/log/job-output.txt you'll see that currently when running the sanity tests it's currently running the import and compile tests against all versions of python (because the tests are running inside of docker)

Because Amazon have dropped support for Python < 3.6 in the SDK we depend on ( https://aws.amazon.com/blogs/developer/announcing-the-end-of-support-for-python-3-4-and-3-5-in-the-aws-sdk-for-python-and-aws-cli-v1/ ) we are dropping support for the old versions of Python in our 2.0.0 release ( https://github.com/ansible-collections/amazon.aws/pull/369/files )

This in turn means we don't care about the Python 2.6, 2.7, 3.4 or 3.5 compile/import tests

pabelanger:
@tremble When I last looked at this, the issue is ansible-test via docker doesn't not expose a way to toggle this. Like we can do via virtualenvs, and --python flag.

What I think you should be doing, is running ansible-test sanity via network-ee, like we do for other collections. That will default to just python38 for now:

https://dashboard.zuul.ansible.com/t/ansible/build/9e4c7663dadd4705b0a74b103757252a/log/job-output.txt#569

Then you can drop the docker sanity images here. If you want to keep using them, we need to open an issue via ansible-core to allow for users to only select specific versions of python

pabelanger:
I reopened https://github.com/ansible-collections/amazon.aws/pull/353 to demo what I mean

tremble:
`ansible-test sanity --python 3.8 --docker` runs fine for me locally and limits the tests to a specific version of python

roles/ansible-test/tasks/init_test_options.yaml
```diff
@@ -49,7 +49,7 @@
     ansible_test_options: "{{ ansible_test_options }} --python {{ ansible_test_python }}"
   when:
     - ansible_test_python
-    - not ansible_test_docker
+    - ansible_test_docker_python or (not ansible_test_docker)
```
pabelanger, line 52: If this is the case, you can just remove this check now, and we'll pass the --python option to the container.  However, when I first set it up this didn't work.

see: https://github.com/ansible/ansible-zuul-jobs/pull/776 for the original support

roles/ansible-test/defaults/main.yaml
```diff
@@ -14,6 +14,7 @@ ansible_test_docker: false
 # Default ansible-test options
 ansible_test_git_branch: 'devel'
 ansible_test_python: "{{ ansible_python_version.split('.')[:2]|join('.') }}"
+ansible_test_docker_python: false
```
pabelanger, line 17: this shouldn't be needed, we can just remove it and set ansible_test_python with the change below

roles/ansible-test/tasks/init_test_options.yaml
tremble: Ok, that works for me, I'd assumed there was a reason why a different collection didn't want to run with the limits

tremble:
@pabelanger Anything more you'd need to get this merged?

PR #1041, opened 2021-08-13
Only split the aws work over 6 jobs

pabelanger:
The aws semaphore is capped at 6, no need trying to use more resources
then that.

### DIFF
--- a/roles/ansible-test-splitter/files/split_targets.py
+++ b/roles/ansible-test-splitter/files/split_targets.py
@@ -9,7 +9,8 @@ if len(sys.argv) == 3:
     targets_from_cli = sys.argv[2].split(" ")
 else:
     targets_from_cli = []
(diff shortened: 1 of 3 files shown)


tremble:
The advantage of running in multiple groups even if the semaphore limits us to 6, is that no single job is longer than 60 minutes, at which point the temporary AWS credentials expire.

pabelanger:
recheck

pabelanger:
Yes, that is true. The real issue here, is the extra 5 / 10mins that add up for each node to boot. At worst case, this is an extra 60mins or so to merge time. if we are waiting on a lock to free up.  Where if we cap any 1 job, to only run 6 tests, it is possible to hit the expire.

Lets see what happens, and we can revert this.

The better thing here, is to get more aws credentials and longer expire, which I don't know who manages that.

PR #1042, opened 2021-08-13
Remove ansible-test gate jobs

pabelanger:
For some reason these have been deleted in check but not gate. Some one
needs to dig more into this.


pabelanger:
@goneri when you are back, fell free to revert this. I'm not sure why check jobs are missing

PR #1050, opened 2021-08-14
Migrate ansible-tox-* jobs to centos-8-stream

pabelanger:
This should result in some faster job runtimes on newer flavors.


pabelanger:
recheck

PR #1063, opened 2021-08-17
[kubernetes.core] Running molecule testing using tox environment

abikouo:
This pull request adds new features to the ``Kubernetes.core`` CI: 
- Running molecule using tox 
- Molecule testing over ansible 2.9 and 2.10 releases

### DIFF
--- a/playbooks/ansible-cloud/k8s/integration.yaml
+++ b/playbooks/ansible-cloud/k8s/integration.yaml
@@ -3,6 +3,15 @@
   vars:
     ansible_zuul_src_dir: "{{ ansible_user_dir }}/{{ zuul.projects['github.com/ansible/ansible-zuul-jobs'].src_dir }}"
   tasks:
+    - name: Install molecule and collection dependencies
+      pip:
+        name:
+          - yamllint
+          - kubernetes
+          - flake8
+        virtualenv: '{{ ansible_test_venv_path }}'
+        virtualenv_python: python3
+
     - name: Integration tests
       shell: "source {{ ansible_test_venv_path }}/bin/activate; make test-integration"
       args:
--- a/roles/molecule-test-splitter/library/molecule_splitter.py
+++ b/roles/molecule-test-splitter/library/molecule_splitter.py
@@ -1,0 +1,125 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2021, Aubin Bikouo <@abikouo>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+
+__metaclass__ = type
+
+
+DOCUMENTATION = r'''
+
+module: molecule_splitter
+
+short_description: Split molecule scenario to be executed on zuul jobs.
+
+author:
+    - "Aubin Bikouo (@abikouo)"
+
+description:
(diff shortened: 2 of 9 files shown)


playbooks/ansible-cloud/k8s/integration.yaml
```diff
@@ -3,6 +3,15 @@
   vars:
     ansible_zuul_src_dir: "{{ ansible_user_dir }}/{{ zuul.projects['github.com/ansible/ansible-zuul-jobs'].src_dir }}"
   tasks:
+    - name: Install molecule and collection dependencies
```
pabelanger, line 6: we should move these into the test-requirements.txt file

roles/molecule-test-splitter/library/molecule_splitter.py
```diff
@@ -0,0 +1,125 @@
+#!/usr/bin/python
```
pabelanger, line 1: I'd prefer not to create a 2nd splitter role. I would much rather see us refactor ansible-test-splitter into something more generic.

zuul.d/project-templates.yaml
```diff
@@ -663,7 +663,19 @@
...
+        - ansible-test-molecule-kubernetes-core_devel_2
+        - ansible-test-molecule-kubernetes-core_devel_3
+        - ansible-test-molecule-kubernetes-core_2.9_0
+        - ansible-test-molecule-kubernetes-core_2.9_1
+        - ansible-test-molecule-kubernetes-core_2.9_2
+        - ansible-test-molecule-kubernetes-core_2.9_3
+        - ansible-test-molecule-kubernetes-core_2.10_0
+        - ansible-test-molecule-kubernetes-core_2.10_1
+        - ansible-test-molecule-kubernetes-core_2.10_2
+        - ansible-test-molecule-kubernetes-core_2.10_3
```
pabelanger, line 678: to be honest, I doubt you need to do this.  What we should do, is create a 2nd nodeset for k8s creating a controller / appliance model. Then, see how long the job runtimes are.

playbooks/ansible-cloud/k8s/integration.yaml
abikouo: this is temporary, integration tests will be removed

roles/molecule-test-splitter/library/molecule_splitter.py
abikouo: trying to merge this splitter with the amazon one will increase the complexity of the role, this is just a simple executable listing all available scenario.

zuul.d/project-templates.yaml
abikouo: We definitely need it, otherwise, the molecule testing is too long, I try using a controller and appliance and it does not reduce, the kubernetes collection is using the wait option in many test case, this increase the total elapsed time.

pabelanger: what is too long, can you share job runtimes? Rather then dynamically splitting them, and maintaining that logic we can just make it static.

roles/molecule-test-splitter/library/molecule_splitter.py
pabelanger: yes, that is true. However, highlights exactly why we should use a single code base.

We really shouldn't need to dynamically create child jobs all the time, lets get into a 1:1 and sync on this to better understand what troubles you are dealing with.

PR #1066, opened 2021-08-19
ansible.network  meta collection network-ee tests added

rohitthakur2590:
Signed-off-by: Rohit Thakur <rohitthakur2590@outlook.com>
Depends-On: https://github.com/ansible/network-ee/pull/119

### DIFF
--- a/zuul.d/ansible-network-jobs.yaml
+++ b/zuul.d/ansible-network-jobs.yaml
@@ -1,6 +1,19 @@
 ---
 
 - job:
+    name: ansible-network-ios-appliance
+    parent: ansible-network-appliance-base
+    pre-run: playbooks/ansible-network-ios-appliance/pre.yaml
+    run: playbooks/ansible-network-ios-appliance/run.yaml
+    host-vars:
+      ios-15.6-2T:
+        ansible_network_os: ios
+    required-projects:
+      - name: github.com/ansible/ansible-zuul-jobs
+    nodeset: ios-15.6-2T-python38
+    timeout: 10800
+
+- job:
     name: network-ee-integration-tests-ansible-network
     parent: ansible-network-ios-appliance
 
@@ -8,3 +21,10 @@
     name: network-ee-integration-tests-ansible-network
     parent: network-ee-integration-tests
     nodeset: ios-15.6-2T-python38
+
+- job:
+    name: network-ee-integration-tests-ios-libssh
+    parent: network-ee-integration-tests-ansible-network
+    vars:
+      test_ansible_network_cli_ssh_type: libssh
+      test_ansible_skip_tags: local
--- a/zuul.d/ansible-network-jobs.yaml
+++ b/zuul.d/ansible-network-jobs.yaml
@@ -1,0 +1,10 @@
+---
+
+- job:
+    name: network-ee-integration-tests-ansible-network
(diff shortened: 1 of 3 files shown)


zuul.d/ansible-network-jobs.yaml
```diff
@@ -0,0 +1,10 @@
+---
+
+- job:
+    name: network-ee-integration-tests-resource_manager-ios
```
pabelanger, line 4: this job should be calle

network-ee-integration-tests-ansible-network

so we reference the collection, not the target

zuul.d/ansible-network-jobs.yaml
```diff
@@ -0,0 +1,10 @@
+---
+
+- job:
+    name: network-ee-integration-tests-ansible-network
+    parent: ansible-network-ios-appliance
```
pabelanger, line 5: This isn't correct, if we parent to ios, then all testing will use this image.  Do you plan to run tests against eos for example?

roles/ansible-test-inventory/templates/inventory-ios.j2
```diff
@@ -4,6 +4,9 @@
 [ios]
 ios-15.6-2T ansible_port=22 ansible_host={{ hostvars['ios-15.6-2T'].ansible_host }} ansible_user=zuul
 
+[resource:children]
+ios
```
pabelanger, line 8: this should not be needed

zuul.d/ansible-network-jobs.yaml
rohitthakur2590: > This isn't correct, if we parent to ios, then all testing will use this image. Do you plan to run tests against eos for example?

yes we might want to add tests for multiple platforms in future

roles/ansible-test-inventory/templates/inventory-ios.j2
rohitthakur2590: So before doing this we were getting consistent failures because of missing inventory.

pabelanger: this was used by venv testing, no longer with network-ee.  If you depends-on https://github.com/ansible-collections/cisco.ios/pull/397 you'll see we actually delete it now.

zuul.d/ansible-network-jobs.yaml
pabelanger: The way this is done now, will not work as you expect. I would like you to review and even try to land https://github.com/ansible-collections/cisco.ios/pull/397

This will give you a better understanding of how the new network-ee jobs work.

Basically, this job would just run cisco.ios tests, not the ansible.network collection.  For that to happen, we need to setup collection_name / collection_namespace in the inventory file

rohitthakur2590: @pabelanger so are you suggesting having a specific job like `ansible-test-network-integration-ios-python36`
because for now, we are testing for ios.

roles/ansible-test-inventory/templates/inventory-ios.j2
rohitthakur2590: okay I'll try with it .

rohitthakur2590:
recheck

PR #1071, opened 2021-08-23
Add aws-complete pipeline which runs all integration tests in collection

jillr:
Depends-On: https://github.com/ansible/project-config/pull/924


jillr:
recheck

jillr:
We've tested this on both AWS collections and everything seems to be in order.

goneri:
This is really cool, I think we will soon have something similar for vmware.

goneri (approved):

PR #1073, opened 2021-08-24
ansible-tox-py37: ensure py37 is installed

goneri:
Ensure we've got the python3.7 package installed for the case where it's
not in the base system (AWS).


pabelanger:
what is still using ansible-tox-py37?  We likely should migrate things to atleast py38

goneri:
> ansible-tox-py37

It's pulled by `ansible-python3-jobs`.

pabelanger:
we should update that. Or deprecate it

PR #1091, opened 2021-09-02
Ensure splitter script runs all requested targets

gravesm:
This fixes a couple problems with the splitter script. The first problem
is that modules listed in aliases files were not being run. The second
problem is that non-slow targets may be silently skipped in cases when
there are six or more slow jobs being run.

### DIFF
--- a/roles/ansible-test-splitter/files/split_targets.py
+++ b/roles/ansible-test-splitter/files/split_targets.py
@@ -15,7 +15,6 @@ args = parser.parse_args()
 batches = []
 # NOTE(pabelanger): Hardcode this to 6 because that is the semaphore in zuul.
 total_jobs = 6
-slow_targets = []
 regular_targets = []
 
 jobs = [f"{args.prefix}{i}" for i in range(total_jobs)]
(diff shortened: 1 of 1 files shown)


roles/ansible-test-splitter/files/split_targets.py
```diff
@@ -28,9 +27,9 @@
         continue
     if not aliases.is_file():
         continue
-    if targets_from_cli and target.name not in targets_from_cli:
-        continue
     lines = aliases.read_text().split("\n")
+    if targets_from_cli and not {*lines, target.name} & {*targets_from_cli}:
```
goneri, line 31: I would suggest to write `set(targets_from_cli)` instead of `{*targets_from_cli}`. This just because I'm more familiar with this syntax.

roles/ansible-test-splitter/files/split_targets.py
```diff
@@ -28,9 +27,9 @@
         continue
     if not aliases.is_file():
         continue
-    if targets_from_cli and target.name not in targets_from_cli:
-        continue
     lines = aliases.read_text().split("\n")
+    if targets_from_cli and not {*lines, target.name} & set(targets_from_cli):
```
goneri, line 31: Can you do the same for the first one :-)?

gravesm: No, the `set()` function only accepts a single argument. This is why I used the other syntax. I could create a new list and use that, I guess, though.

goneri: ah I see.

goneri (commented):
Follow-up to a quick sync with @gravesm. We land this PR, and I will work on a set of unit-test to clarify the expected behaviours. It will simplify the long term maintenance.

PR #1093, opened 2021-09-02
tox: run pytest

goneri:
Add a pytest environment and run it against the role directory. The goal
is to add unit-test to `ansible-test-splitter` and this is the first step.


zuul.d/project.yaml
```diff
@@ -4,6 +4,7 @@
       - ansible-python-jobs
     check:
       jobs:
+        - ansible-tox-py36
```
pabelanger, line 7: lets flip to ansible-tox-py38 given this runs under python38

tox.ini
```diff
@@ -19,3 +19,11 @@ commands =
 setenv =
   ANSIBLE_CALLBACK_PLUGINS = {envsitepackagesdir}/ara/plugins/callbacks
 commands = {posargs}
+
+[testenv:pytest]
+deps =
+  ansible
```
pabelanger, line 25: nit, maybe change this to ansible-core, to minimize the install. But unsure what deps really look like

goneri: Oh +1, I always forget about that.

zuul.d/project.yaml
```diff
@@ -4,6 +4,7 @@
       - ansible-python-jobs
     check:
       jobs:
+        - ansible-tox-py38
```
pabelanger, line 7: this ran tox -epy38 not tox -epytest. You'll need to set tox_envlist here

pabelanger (changes requested):

PR #1095, opened 2021-09-03
add test coverage for split_targets.py

goneri:
Depends-On: https://github.com/ansible/ansible-zuul-jobs/pull/1093

### DIFF
--- a/roles/ansible-test-splitter/files/split_targets.py
+++ b/roles/ansible-test-splitter/files/split_targets.py
@@ -6,52 +6,116 @@ import json
 import argparse
 import os
 
-parser = argparse.ArgumentParser()
-parser.add_argument('-t', '--targets', help="list of targets for integration testing. example: 'ec2_tag ami_info ec2_instance'",default="")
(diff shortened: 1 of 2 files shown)


goneri:
@gravesm ^

roles/ansible-test-splitter/files/split_targets.py
```diff
@@ -6,52 +6,116 @@
...
+        batch = regular_targets[x::remaining_jobs]
+        if batch:
+            batches.append(batch)
+    return batches
+
+
+def build_result_struct(jobs, batches):
+    result = {
+        "data": {
+            "zuul": {"child_jobs": jobs[0 : len(batches)]},
```
gravesm, line 81: What happens if `len(batches)` is longer than `len(jobs)`? Would that mean some targets don't get run?

goneri: Yes, probably. The goal of this PR is to add test coverage. I suggest we merge this PR and start improving stuff. It will be easier to understand what we do.

gravesm (approved):

PR #1096, opened 2021-09-03
split_targets: handle target redirection

goneri:
Evalate the target redirection and use a flatten list internally. Take the `--targets` parameter into consideration during the operation.

### DIFF
--- a/roles/ansible-test-splitter/files/split_targets.py
+++ b/roles/ansible-test-splitter/files/split_targets.py
@@ -30,16 +30,39 @@ def is_slow(lines):
     return False
 
 
-def iter_target_dir(collection_path):
-    targets = PosixPath(os.path.join(collection_path, "tests/integration/targets/"))
-    for target in targets.glob("*"):
+def get_all_targets(collection_path):
+    """Return all the targets from a directory."""
+    targets_dir = PosixPath(os.path.join(collection_path, "tests/integration/targets/"))
+    targets = {}
+    for target in targets_dir.glob("*"):
         aliases = target / "aliases"
         if not target.is_dir():
             continue
         if not aliases.is_file():
             continue
         lines = aliases.read_text().split("\n")
-        yield (target, lines)
+        targets[target.name] = lines
+    return targets
+
+
+def evaluate_redirection(in_targets, args_targets):
+    """Take get_all_targets() output and evaluate the redirection of the aliases files."""
+    targets = {}
+    allowed_targets = args_targets or [*in_targets.keys()]
+    for target_name, lines in in_targets.items():
+        for l in lines:
+            if l not in allowed_targets:
+                continue
+            if l not in in_targets:
+                continue
+
+            if l not in targets:
+                targets[l] = in_targets[l]
+            break
+        else:
+            if target_name in allowed_targets:
+                targets[target_name] = lines
+    return targets
 
 
 def to_skip_because_of_targets_parameters(target_name, lines, targets_from_cli):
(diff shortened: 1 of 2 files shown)


goneri:
@gravesm ^

roles/ansible-test-splitter/files/split_targets.py
```diff
@@ -30,16 +30,35 @@ def is_slow(lines):
...
             continue
         if not aliases.is_file():
             continue
         lines = aliases.read_text().split("\n")
-        yield (target, lines)
+        targets[target.name] = lines
+    return targets
+
+
+def evaluate_redirection(in_targets):
```
goneri, line 48: Actually, we should take `args.targets` as a parameter and use the value to decide what target we keep.

goneri:
@tremble ^

roles/ansible-test-splitter/files/test_split_targets.py
```diff
@@ -34,6 +59,42 @@ def test_is_slow():
...
+    }
+
+
+def test_evaluate_redirection():
+    _in = {
+        "t1": ["cloud/aws", "slow"],
+        "t2": ["t1"],
+        "t3": ["cloud/aws", "", "disabled"],
+    }
+    out = {
```
gravesm, line 77: I don't think I understand what evaluate_redirection is supposed to be doing. Why aren't we including t2 in the results? Isn't that going to lead to targets being skipped that shouldn't be?

tremble: I'm as confused as @gravesm.

Is this backwards...
If t1 is an alias of t2, then the t2 tests should be running if t1 or t2 is specified as the target...

goneri: `t2` is an alias for `t1`. We include `t1` in the result. If we also include `t2`, it's like having two times `t1`. And we may dschedule t1 on two different jobs.

tremble: The input is the equivalent of:
t1/aliases:
```
cloud/aws
slow
```

t2/aliases:
```
t1
```

t3/aliases:
```
cloud/aws
disabled
```

Which is a little strange...  What's more common would be

t2/aliases:
```
t4
```

And then args_targets (below) containing "t4" rather than "t1", "t2" or "t3".

goneri: Thanks. I was reading the redirection declaration in the wrong way.

```
<tremble> goneri, I'm a little confused as to what usecase ansible/ansible-zuul-jobs#1096 is trying to cover, it feels like your test case is trying to cover a use-case we never see, but I may be misunderstanding...
<goneri> tremble, I was looking for an example and I think it helps me to understand my mistake.
<goneri> here, ec2_eni_info is an alias for ec2_eni: https://github.com/ansible-collections/amazon.aws/blob/main/tests/integration/targets/ec2_eni/aliases
<goneri> and I misunderstood how the aliases are defined. I thought there where in the opposite direction. Something like: echo ec2_eni > tests/integration/targets/ec2_eni_info/aliases
```

goneri: I will close this PR and reopen a new one with no code change and just the valid unit-tests.

gravesm: With the following alias file: https://github.com/ansible-collections/community.aws/blob/main/tests/integration/targets/aws_config/aliases
```
$ python split_targets.py -c /home/mgraves/git/ansible_collections/community/aws/ -t aws_config_rule
{"data": {"zuul": {"child_jobs": []}, "child": {"targets_to_test": []}}}
```
I would expect this to include `aws_config` in the targets to test, right?

goneri: Yes

> With the following alias file: https://github.com/ansible-collections/community.aws/blob/main/tests/integration/targets/aws_config/aliases
> 
> ```
> $ python split_targets.py -c /home/mgraves/git/ansible_collections/community/aws/ -t aws_config_rule
> {"data": {"zuul": {"child_jobs": []}, "child": {"targets_to_test": []}}}
> ```
> 
> I would expect this to include `aws_config` in the targets to test, right?

Yes. This is what I misunderstood.

goneri:
As decided, this is the new PR with only the unit-tests: https://github.com/ansible/ansible-zuul-jobs/pull/1103

PR #1097, opened 2021-09-06
Slow targets should only be wrapped in a list once

tremble:
We're currently seeing 'slow' tests double-wrapped in a list, this is breaking things:

```
  "for_zuul_return": {
    "data": {
      "child": {
        "targets_to_test": [
          [   
            [   
              "ec2_vpc_subnet"
            ]   
          ],  
          [   
            [   
              "ec2_vol"
            ]   
          ],  
          [   
            [   
              "ec2_ami"
            ]   
          ],  
          [   
            [   
              "ec2_eni"
            ]   
          ],  
          [   
            "aws_caller_info",
            "aws_s3",
            "lookup_aws_account_attribute",
            "ec2_elb_lb",
            "ec2_vpc_igw",
            "ec2",
            "inventory_aws_ec2",
            "module_utils_core",
            "ec2_key",
            "s3_bucket",
            "cloudformation",
            "ec2_vpc_dhcp_option",
            "module_utils_waiter"
          ],  
          [
            "ec2_metadata_facts",
            "ec2_spot_instance",
            "aws_az_info",
            "ec2_vpc_endpoint_service_info",
            "ec2_vpc_route_table",
            "inventory_aws_rds",
            "ec2_tag",
            "ec2_vpc_nat_gateway",
            "ec2_vpc_net",
            "lookup_aws_secret",
            "lookup_aws_service_ip_ranges",
            "module_utils_ec2"
          ]
        ]
      },
      "zuul": {
        "child_jobs": [
          "ansible-test-cloud-integration-aws-py36_0",
          "ansible-test-cloud-integration-aws-py36_1",
          "ansible-test-cloud-integration-aws-py36_2",
          "ansible-test-cloud-integration-aws-py36_3",
          "ansible-test-cloud-integration-aws-py36_4",
          "ansible-test-cloud-integration-aws-py36_5"
        ]
      }
    }
  }
}
```

PR #1098, opened 2021-09-07
Fix splitter to correctly format list of jobs

gravesm:
The slow jobs get placed into separate lists later down in `build_up_batches()`.


gravesm:
Closing because duplicate of https://github.com/ansible/ansible-zuul-jobs/pull/1097

PR #1111, opened 2021-09-13
add periodical jobs to satisfy ansible inclusion

goneri:
Schedule periodic jobs in order to fulfil ansible-inclusion
requirements. Collections are:

- vmware.vmware_rest
- cloud.common


goneri:
@jillr ^

PR #1114, opened 2021-09-14
cloud.common/vmware_rest: periodic jobs need the artifacts

goneri:
The sanity jobs need the collection artifacts. This commit ensures we
call `build-ansible-collection` first.

### DIFF
--- a/zuul.d/project-templates.yaml
+++ b/zuul.d/project-templates.yaml
@@ -495,6 +495,7 @@
         - ansible-test-sanity-docker-stable-2.10
         - ansible-test-sanity-docker-stable-2.11
         - ansible-test-units-cloud-common-python38
+        - build-ansible-collection
 
 - project-template:
     name: ansible-collections-community-vmware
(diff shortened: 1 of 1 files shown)


zuul.d/project-templates.yaml
```diff
@@ -457,6 +457,7 @@
         - ansible-test-sanity-docker-stable-2.10
         - ansible-test-sanity-docker-stable-2.11
         - ansible-test-units-cloud-common-python38
+        - build-ansible-collection
```
pabelanger, line 460: this won't actually build artifacts, as there are no projects to build.

goneri: I just need the cloud.common artifact. I thought this was enough for this case.

goneri: This is what is actually done in the `publish-to-galaxy-master-only` template.

pabelanger: yes, that is correct. However that would use the released artifacts on galaxy.a.c. however, if you are doing periodic testing, you'd want to ensure the latest code from git is used likely, not released?

goneri: I would like to let build-ansible-collection prepare the tarball of the collection. This way the jobs can run properly.

PR #1119, opened 2021-09-20
Pin OKD to specific k8s.core version

gravesm:
The version here needs to be able to fulfill what's in OKD's galaxy.yml
file.

### DIFF
--- a/zuul.d/project-templates.yaml
+++ b/zuul.d/project-templates.yaml
@@ -457,6 +457,7 @@
         - build-ansible-collection:
             required-projects:
               - name: github.com/ansible-collections/kubernetes.core
+                override-checkout: 2.1.1
         - ansible-test-sanity-docker
         - ansible-test-sanity-docker-stable-2.9
         - ansible-test-sanity-docker-stable-2.10
--- a/zuul.d/project-templates.yaml
+++ b/zuul.d/project-templates.yaml
@@ -457,7 +457,7 @@
         - build-ansible-collection:
             required-projects:
               - name: github.com/ansible-collections/kubernetes.core
-                override-checkout: 2.1.1
+                override-checkout: stable-2.1
         - ansible-test-sanity-docker
         - ansible-test-sanity-docker-stable-2.9
         - ansible-test-sanity-docker-stable-2.10


abikouo (approved):

zuul.d/project-templates.yaml
```diff
@@ -457,6 +457,7 @@
         - build-ansible-collection:
             required-projects:
               - name: github.com/ansible-collections/kubernetes.core
+                override-checkout: 2.1.1
```
goneri, line 460: This looks wrong to me, because we won't test any future 2.1.x versions. Can we instead branch 2.1.x and use this branch?

goneri (changes requested):

zuul.d/project-templates.yaml
goneri: Overall, what prevents us from using master here? I think this PR is connected with yours https://github.com/openshift/community.okd/pull/109, would this be a better fix?

gravesm: The problem with using openshift/community.okd#109 to fix this is that we are then forced to always keep OKD working with the current release of k8s.core. While we try to do this, there is often a lag. If we do a major release of kubernetes.core then our okd CI would be broken until we can fix it. Ideally, the voting CI jobs would use galaxy.yml for installing dependencies, but since it doesn't, I'd rather we pin the kubernetes.core version here.

goneri: I just pushed a fix that address a breaking change that should have been caught by the CI https://github.com/ansible-collections/kubernetes.core/pull/239
This is the kind of regression that we can identify before we introduce them if we do cross-project testing. And we cannot do cross-project testing properly if we rely on the galaxy.yml, because it will fetch the last stable version.

goneri:
The release of OKD are difficult because we don't do any cross-project testing, so any commit on kubernetes.core can silently break OKD and we don't know until we do the release and try to upgrade OKD. And it's already months to late. The situation will remain if we continue to pin OKD to the stable branch. This is the main problem that we need to address. This being said, I applied the gate label.

goneri (approved):

PR #1121, opened 2021-09-20
Add publish-to-automation-hub-3pci

alinabuzachis:
Add `publish-to-automation-hub-3pci` to be used by `community.okd `

### DIFF
--- a/zuul.d/project-templates.yaml
+++ b/zuul.d/project-templates.yaml
@@ -1393,6 +1393,25 @@
                 soft: true
 
 - project-template:
+    name: publish-to-automation-hub-3pci
+    description: |
+      Publish an Ansible collection to Ansible Automation Hub
+    third-party-check:
+      jobs:
+        - build-ansible-collection
+    pre-release:
+      jobs:
+        - release-ansible-collection-automation-hub
+    release:
+      jobs:
+        - release-ansible-collection-automation-hub
+        - release-ansible-collection-announcement:
+            dependencies:
+              - name: release-ansible-collection-automation-hub
+              - name: release-ansible-collection-galaxy
+                soft: true
+
+- project-template:
     name: publish-to-galaxy
     description: |
       Publish an Ansible collection to Ansible Galaxy


goneri (approved):

pabelanger:
@alinabuzachis @goneri we don't upload community collections to automation hub, so curious why we are doing this?

zuul.d/project-templates.yaml
```diff
@@ -1392,6 +1392,25 @@
               - name: release-ansible-collection-galaxy
                 soft: true
 
+- project-template:
+    name: publish-to-automation-hub-3pci
+    description: |
+      Publish an Ansible collection to Ansible Automation Hub
+    third-party-check:
```
pabelanger, line 1399: this is the wrong pipeline for 3pci-check

alinabuzachis:
> @alinabuzachis @goneri we don't upload community collections to automation hub, so curious why we are doing this?

`redhat.openshift` is cut from `community.okd` and we publish it to the automation hub. @goneri

pabelanger:
@alinabuzachis how does that work? If you were to tag community.okd today, it would build that tarball and publish it to automation hub.  Where does the redhat.openshift changes take place?

zuul.d/project-templates.yaml
goneri: oh indeed. Let's fix that.

goneri: https://github.com/ansible/ansible-zuul-jobs/pull/1124

pabelanger:
what we should do, is the same thing we do for vmware modules. i believe @goneri will sync commits between the 2 projects, which means 2 repos.  I don't think we have automation that will take the community repo directly and rename it to a redhat collection.

goneri:
> > @alinabuzachis @goneri we don't upload community collections to automation hub, so curious why we are doing this?
> 
> @pabelanger `redhat.openshift` is cut from `community.okd` and we publish it to the automation hub. @goneri

Oh! Yes, you're right. I totally forgot about.

goneri:
> > > @alinabuzachis @goneri we don't upload community collections to automation hub, so curious why we are doing this?
> > 
> > 
> > @pabelanger `redhat.openshift` is cut from `community.okd` and we publish it to the automation hub. @goneri
> 
> Oh! Yes, you're right. I totally forgot about.

-> https://github.com/ansible/project-config/pull/947

PR #1122, opened 2021-09-21
vmware: Remove Python 3.6 unit tests

Akasurde:


goneri:
I believe we still support 3.6, don't we?

PR #1132, opened 2021-09-22
Use Python 3.8 for NX-OS devel jobs

NilashishC:


zuul.d/cisco-nxos-jobs.yaml
```diff
@@ -71,9 +71,9 @@
 - job:
     name: ansible-test-network-integration-nxos-cli-python36
```
GomathiselviS, line 72:
```suggestion
    name: ansible-test-network-integration-nxos-cli-python38
```

GomathiselviS: After changing here, you might have to change the job name in project-template.

PR #1129, opened 2021-09-22
[WIP]  TESTING: migrate to network ee

rohitthakur2590:
Signed-off-by: Rohit Thakur <rohitthakur2590@outlook.com>
Depends-On: https://github.com/ansible/network-ee/pull/103

PR #1087, opened 2021-09-01
IOSXR: Migrate integration tests jobs to network EE jobs.

ashwini-mhatre:
Depends-on: https://github.com/ansible/network-ee/pull/122

PR #1139, opened 2021-09-24
Move netcommon-specific test runs to python38

Qalthos:
The underlying jobs have moved, so this should work now

This is obsoleted by https://github.com/ansible/ansible-zuul-jobs/pull/1087 and https://github.com/ansible/ansible-zuul-jobs/pull/1129, but unblocks testing in the meantime


GomathiselviS (approved):
LGTM

PR #1157, opened 2021-10-06
Add fedora-34 for ansible-tox-linters 

GomathiselviS:
Replace centos8-stream node with fedora-34.

The centos-8-stream images are available on OpenStack only. As a result, we cannot scale our workload on AWS.


pabelanger:
@GomathiselviS  this efficiently only now runs ansible-tox-linter jobs under aws, which is a single pool vs 4 regions.  I'd like to revert this until all regions have the same image

PR #1171, opened 2021-10-08
ansible-test-sanity-docker: test milestone

goneri:
- test the new milestone branch
- continue to test devel but with voting==false
- ensure we use podman when it's relevant

### DIFF
--- a/zuul.d/ansible-test-jobs.yaml
+++ b/zuul.d/ansible-test-jobs.yaml
@@ -23,6 +23,21 @@
       container_command: podman
 
 - job:
+    name: ansible-test-sanity-docker-devel
+    parent: ansible-test-sanity-docker
+    required-projects:
+      - name: github.com/ansible/ansible
+        override-checkout: devel
+    voting: false
+
+- job:
+    name: ansible-test-sanity-docker-milestone
+    parent: ansible-test-sanity-docker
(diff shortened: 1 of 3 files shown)


zuul.d/ansible-test-jobs.yaml
```diff
@@ -22,6 +22,21 @@
       ansible_test_python: 3.8
       container_command: podman
 
+- job:
+    name: ansible-test-sanity-docker-devel
+    parent: ansible-test-sanity-docker
+    required-projects:
+      - name: github.com/ansible/ansible
+        override-checkout: devel
+    voting: false
```
pabelanger, line 31: we shouldn't do this my default, no one looks at non-voting jobs and it is a waste of resources.  We should keep it voting and projects that don't want to track devel can just remove it now IMO.

This however expose the issue where we are no longer testing against the latest code and makes it more harder to get into compliance closer to release time.

goneri: If we stop testing `devel`, we will end-up with breaking change in milestone. If we keep `devel` voting, we end up with broken tests on our side. I dislike both. But at this stage, I would prefer to continue to test devel as non-voting, and see if we can report the issues efficiently.

PR #1201, opened 2021-10-25
[python3.9] Add ansible-tox-py39 job

goldyfruit:
`molecule-libvirt`'s CI keep failing because `python38-devel` is not available in Fedora 34.

Python 3.9 is the default version in Fedora 34

_EDIT_

From @goneri comment about possible confusion between CentOS and Fedora:

> Yeah in our tests we are using `ansible-tox-py38` which is based on `centos-8-stream` for Fedora :D
> https://github.com/ansible-community/molecule-libvirt/blob/main/zuul.d/layout.yaml#L4

### DIFF
--- a/playbooks/ansible-tox-py39/pre.yaml
+++ b/playbooks/ansible-tox-py39/pre.yaml
@@ -1,0 +1,8 @@
+---
+- hosts: all
+  tasks:
+    - name: Ensure python3.9 is present
+      become: true
+      package:
+        name: python39-devel
+        state: present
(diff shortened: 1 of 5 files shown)


goldyfruit:
@goneri all good for you?

goneri (approved):

goneri:
Looks good to me. The PR description is a bit confusing since you actually since the change is more about Centos8.

goldyfruit:
> Looks good to me. The PR description is a bit confusing since you actually since the change is more about Centos8.

I edited my initial description with the explanation.

PR #1203, opened 2021-11-02
split_targets: Allow to run unstable test

Akasurde:
When targets are specified explicitly, run targets
even if they are marked as unstable.

### DIFF
--- a/roles/ansible-test-splitter/files/split_targets.py
+++ b/roles/ansible-test-splitter/files/split_targets.py
@@ -14,18 +14,20 @@ def get_job_list(prefix, total_jobs):
     return [f"{prefix}{i}" for i in range(total_jobs)]
 
 
-def to_skip_because_disabled(lines):
-    if "disabled" in lines:
-        return True
+def to_skip_because_unstable(lines):
     if "unstable" in lines:
         return True
     return False
 
 
-def is_slow(lines):
-    if "slow" in lines:
+def to_skip_because_disabled(lines):
+    if "disabled" in lines:
         return True
-    if "# reason: slow" in lines:
+    return False
+
+
+def is_slow(lines):
+    if "slow" in lines or "# reason: slow" in lines:
         return True
     return False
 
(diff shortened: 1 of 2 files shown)


Akasurde:
cc @goneri

roles/ansible-test-splitter/files/split_targets.py
```diff
@@ -59,7 +59,7 @@ def get_targets_to_run(targets, targets_from_cli):
     for target_name, lines in targets.items():
         if to_skip_because_of_targets_parameters(target_name, lines, targets_from_cli):
             continue
-        if to_skip_because_disabled(lines):
+        if to_skip_because_disabled(lines) and not targets_from_cli:
```
goneri, line 62: You also include disabled target. Is this something expected? The PR title just mentions the unstable targets.

roles/ansible-test-splitter/files/test_split_targets.py
```diff
@@ -149,3 +150,17 @@ def test_build_up_batches(slow_targets, regular_targets, total_jobs, expected):
 )
 def test_build_result_struct(jobs, batches, expected):
     assert build_result_struct(jobs, batches) == expected
+
+
+@pytest.mark.parametrize(
+    "target,targets_from_cli,expected",
+    [
+        ({"test1": ["cloud/aws"]}, "test1", ([], [])),
+        ({"test1": ["cloud/aws"]}, "", ([], ["test1"])),
```
goneri, line 159: `targets_from_cli` is a list but you pass a string.

roles/ansible-test-splitter/files/split_targets.py
goneri: Also, if `targets_from_cli` is empty, we should ignore the filter. I think we should instead have something like:

`if to_skip_because_unstable(lines) and target_name not in targets_from_cli:`

goneri (changes requested):

goneri (approved):

Akasurde:
@goneri Thanks for the review and merge.

goneri:
> @goneri Thanks for the review and merge.

You're welcome. Thank you for the adjustments.

PR #1204, opened 2021-11-02
Fix check for py39 on fedora 34/35 (old)

ssbarnea:
Address wrongly assumption that py39 devel package is always named `python39-devel`.

Fixes bug introduced by https://github.com/ansible/ansible-zuul-jobs/pull/1201

Required-By: https://github.com/ansible-community/molecule-libvirt/pull/31

### DIFF
--- a/playbooks/ansible-tox-py39/pre.yaml
+++ b/playbooks/ansible-tox-py39/pre.yaml
@@ -1,8 +1,21 @@
 ---
 - hosts: all
+  vars:
+    python3_devel_pkg: python39-devel
   tasks:
-    - name: Ensure python3.9 is present
+
+    - name: Set python package name for Fedora 34
+      set_fact:
+        python3_devel_pkg: python3-devel
+      when: ansible_distribution == 'Fedora' and ansible_distribution_major_version == '34'
+
+    - name: Set python package name for Fedora 35
+      set_fact:
+        python3_devel_pkg: python3.9
+      when: ansible_distribution == 'Fedora' and ansible_distribution_major_version == '35'
+
+    - name: Ensure python3.9 devel is present
       become: true
       package:
-        name: python39-devel
+        name: "{{ python3_devel_pkg }}"
         state: present
--- a/zuul.d/jobs.yaml
+++ b/zuul.d/jobs.yaml
@@ -87,6 +87,11 @@
     nodeset: centos-8-stream
 
 - job:
+    name: ansible-tox-py39-fedora
+    parent: ansible-tox-py39
+    nodeset: fedora-latest-1vcpu
+
+- job:
     name: ansible-build-python-tarball
     description: |
       Test building python tarballs / wheels and the packaging metadata
--- a/zuul.d/project.yaml
+++ b/zuul.d/project.yaml
@@ -10,6 +10,9 @@
         - ansible-tox-py39:
             vars:
               tox_envlist: pytest,black
+        - ansible-tox-py39-fedora:
+            vars:
+              tox_envlist: pytest,black
         - ansible-network-asa-appliance:
             files:
               - playbooks/ansible-network-appliance-base/.*
@@ -69,6 +72,9 @@
             vars:
               tox_envlist: pytest,black
         - ansible-tox-py39:
+            vars:
+              tox_envlist: pytest,black
+        - ansible-tox-py39-fedora:
             vars:
               tox_envlist: pytest,black
         - ansible-network-asa-appliance:


ssbarnea:
@pabelanger @Gomathiselvi-S Anything blocking this fix?

playbooks/ansible-tox-py39/pre.yaml
```diff
@@ -4,5 +4,5 @@
     - name: Ensure python3.9 is present
       become: true
       package:
-        name: python39-devel
+        name: '{% if (ansible_distribution == "Fedora" and ansible_distribution_major_version|int >= 34) %}python-devel{% else %}python39-devel{% endif %}'
```
goneri, line 7: Well, this is pretty complex :-). Please write two different tasks and use a `when` condition to apply the right one depending on the OS.

goneri (changes requested):

playbooks/ansible-tox-py39/pre.yaml
ssbarnea: In fact there is a pattern that I documented few years back which is also used by many roles from zuul-jobs, you can read about it at https://sbarnea.com/ansible/distro-vars/

Still, in this particular case we do have a much better solutions, using a required filename instead of rpm name for installing the dependency and avoiding the change in name.

playbooks/ansible-tox-py39/pre.yaml
```diff
@@ -1,8 +1,22 @@
...
+    - name: Load operating system specific variables
+      include_vars: "{{ item }}"
+      failed_when: false
+      loop:
+          - "family-{{ ansible_os_family | lower }}.yml"
+          - "family-{{ ansible_os_family | lower }}-{{ ansible_distribution_major_version | lower }}.yml"
+          - "{{ ansible_distribution | lower | replace(' ', '-') }}.yml"
+          - "{{ ansible_distribution | lower | replace(' ', '-') }}-{{ ansible_distribution_major_version | lower }}.yml"
+          - "{{ ansible_distribution | lower | replace(' ', '-') }}-{{ ansible_distribution_version.split('.')[0:2] | join('-') | lower }}.yml"
+          - "{{ ansible_distribution | lower | replace(' ', '-') }}-{{ ansible_distribution_version.split('.')[0:3] | join('-') | lower }}.yml"
```
samdoran, line 14: As written, this will include _all_ of these files and error if one does not exists (which is why ignoring errors is necessary as implemented).

Usually the strategy with including distribution specific variables is to include the first found file, from most specific to least specific. You can provide an empty file for `default.yml` or allow it to fail if no files are found.

```suggestion
      include_vars: "{{ lookup('first_found', params) }}"
      vars:
        params:
          files:
            - "{{ ansible_facts.distribution }}-{{ ansible_facts.distribution_version }}.yml"
            - "{{ ansible_facts.os_family }}-{{ ansible_facts.distribution_major_version }}.yml"
            - "{{ ansible_facts.distribution }}.yml"
            - "{{ ansible_facts.os_family }}.yml"
            - default.yml
          paths:
            - vars
```

samdoran (changes requested):
Rename the vars files to avoid needing to transform the fact name:
`family-redhat.yml` --> `RedHat.yml`
`fedora-34.yml` --> `Fedora-34.yml`

And if the package name is `python3-devel` on most distros, create a `default.yml` setting `python3_devel_pkg: python3-devel`, or use the `default` filter in the play.

playbooks/ansible-tox-py39/pre.yaml
ssbarnea: Or instead of debating over subjective nits like lowercase, or order of loading maybe it would just be better to revert the breaking change from https://github.com/ansible/ansible-zuul-jobs/pull/1201?

Clearly that went in without serious testing as it would clearly break on both fedora 34 and 35.

ssbarnea: FYI my approach proves to minimize maintenance for complex setups where you want to only override settings by more specific selectors. With your approach if a new variable needs to be added for all redhat products, you may have to add it to 10 files if that is the number of variances across redhat platforms.

With layered approach, you can minimize the number of places where you define a variable.

playbooks/ansible-tox-py39/pre.yaml
```diff
@@ -1,8 +1,21 @@
...
+        - "{{ ansible_distribution | replace(' ', '-') }}-{{ ansible_distribution_version.split('.')[0:2] | join('-') }}.yml"
+        - "{{ ansible_distribution | replace(' ', '-') }}-{{ ansible_distribution_version.split('.')[0:3] | join('-') }}.yml"
+      tags:
+        - always
+
+    - name: Ensure python3.9 devel is present
       become: true
       package:
-        name: python39-devel
+        name: "{{ python3_devel_pkg | default('python39-devel') }}"
```
pabelanger, line 20: Honestly, this is overly complex and makes is much more difficult to review.  Over using 2 different set_fact statements. I'd like to see us use set_fact and default to python3.9, then check for fedora 34 and change the fact to python39-devel.

Then a single task to install like you have here.

We should make it pretty easier for reviewers here, over optimal code, given the lack of people reviewing code.

ssbarnea: doen

ssbarnea:
> any depends-on to confirm this works?

Yes, there is one at https://github.com/ansible-community/molecule-libvirt/pull/31 which passed after I did an extra recheck.

In fact the reason why I added the extra testing job here was to hopefully prevent similar bug from being introduced when we touch it again.

playbooks/ansible-tox-py39/pre.yaml
```diff
@@ -1,8 +1,21 @@
...
+
+    - name: Set python package name for Fedora 34
+      set_fact:
+        python3_devel_pkg: python3-devel
+      when: ansible_distribution == 'Fedora' and ansible_distribution_major_version == '34'
+
+    - name: Set python package name for Fedora 35
+      set_fact:
+        python3_devel_pkg: python3.9
+      when: ansible_distribution == 'Fedora' and ansible_distribution_major_version == '35'
```
samdoran, line 15: This is far more complicated than using a lookup and gets much harder to read, follow, and update as more distribution/version combinations are added. There is also more overhead and execution time involved since there needs to be one task per distro/version combination instead of one single task that just does a lookup.

```suggestion
    - name: Include distribution specific variables
      include_vars: "{{ lookup('first_found', params) }}"
      vars:
        params:
          files:
            - "{{ ansible_facts.distribution }}-{{ ansible_facts.distribution_version }}.yml"
            - "{{ ansible_facts.os_family }}-{{ ansible_facts.distribution_major_version }}.yml"
            - "{{ ansible_facts.distribution }}.yml"
            - "{{ ansible_facts.os_family }}.yml"
            - default.yml
          paths:
            - vars
```

```yaml
# vars/default.yml
python3_devel_pkg: python39-devel
```

```yaml
# vars/Fedora-34.yml
python3_devel_pkg: python3-devel
```

```yaml
# vars/Fedora-35.yml
python3_devel_pkg: python3.9
```

playbooks/ansible-tox-py39/pre.yaml
```diff
@@ -1,8 +1,21 @@
 ---
 - hosts: all
+  vars:
+    python3_devel_pkg: python39-devel
```
samdoran, line 4: This should be set by the `include_vars` task, not as a play var.

samdoran (changes requested):

PR #1220, opened 2021-11-16
turns network-ee-sanity-tests non-voting

goneri:
The network-ee-sanity-tests job is based on ansible/ansible devel and breaks
pretty often. Then this happens, the check/gate pipelines are blocked and
we've got to resolve the situation ASAP. It's often just a trivial problem
but this still sounds like a source of endless distraction for some.


goneri:
Closing, we're using instead collection specific configuration (https://github.com/ansible-collections/kubernetes.core/pull/284). Thanks @pabelanger for the suggestion.

PR #1221, opened 2021-11-16
use fedora-35 for the security collections

goneri:
Depends-On: https://github.com/ansible/ansible-zuul-jobs/pull/1223

The goal is to remove ubuntu-bionic-1vcpu label from the AWS
configuration. The security collections are the last ones using
it.

Also, remove ansible-security-integration-trendmicro-deepsec-python35
because Fedora 35 does not provide Python 3.5.


goneri:
recheck

PR #1222, opened 2021-11-16
Replace ubuntu-bionic and fedora-34 with fedora-35

GomathiselviS:


goneri:
You may want to Depends-On: https://github.com/ansible/ansible-zuul-jobs/pull/1223 to be sure you've got the right Python version. Also, there is no py35 with Fedora 35.

PR #1236, opened 2021-11-23
replace selinux with selinux-please-lie-to-me on Fedora

goneri:
Use `selinux-please-lie-to-me` on Fedora when Python <= 3.6. These
old Python interpreters are not able to load the selinux lib from
the system (built with Python 3.8 or 3.9).

### DIFF
--- a/playbooks/ansible-test-base/pre.yaml
+++ b/playbooks/ansible-test-base/pre.yaml
@@ -60,8 +60,19 @@
     - name: Create virtualenv for ansible-test
       shell: "virtualenv {{ _virtualenv_options }} ~/venv"
 
+    - name: Set selinux package
+      set_fact:
+        _selinux_package: selinux
+
+    - name: Set selinux package (Fedora)
+      set_fact:
+        _selinux_package: selinux-please-lie-to-me
+      when:
+        - ansible_distribution == 'Fedora'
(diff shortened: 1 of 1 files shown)


playbooks/ansible-test-base/pre.yaml
```diff
@@ -60,8 +60,19 @@
...
+    - name: Set selinux package (Fedora)
+      set_fact:
+        _selinux_package: selinux-please-lie-to-me
+      when:
+        - ansible_distribution == 'Fedora'
+        - ansible_test_python is version('3.6', '<=')
+
     - name: Install selinux into virtualenv
-      shell: ~/venv/bin/pip install selinux "setuptools<50.0.0"
+      shell: '~/venv/bin/pip {{ _selinux_package }} selinux "setuptools<50.0.0"'
```
GomathiselviS, line 75:
```suggestion
      shell: '~/venv/bin/pip install {{ _selinux_package }} selinux "setuptools<50.0.0"'
```

mariolenz:
Thanks @goneri, it looks like this fixes the SELinux issue in the CI. Thanks!

@GomathiselviS Could you please review again? I think the changes you've requested have been implemented.

GomathiselviS (approved):

PR #1252, opened 2021-12-04
Revert change to fedora for ansible-tox-linters

pabelanger:
The change at https://github.com/ansible/ansible-zuul-jobs/pull/1157
basically pins this job to only run under AWS. Where centos-8-stream
available on 4 providers.

I would suggest, project that want to use fedora-latest override the
centos-8-stream nodeset vs applying this to all projects.

### DIFF
--- a/zuul.d/jobs.yaml
+++ b/zuul.d/jobs.yaml
@@ -25,7 +25,7 @@
 - job:
     name: ansible-tox-linters
     parent: tox-linters
-    nodeset: fedora-latest-1vcpu
+    nodeset: centos-8-stream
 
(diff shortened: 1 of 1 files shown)


pabelanger:
@goneri @GomathiselviS this makes the job run under 5mins again. Under aws is was upwards for 10mins, because of all the extra base packages we install and stuff.

I strongly suggest we allocate time to work on the upstream zuul nodepool patch to upload DIBs to aws, so we can finally standardize our images across everything.

pabelanger:
https://review.opendev.org/c/zuul/nodepool/+/735217

PR #1255, opened 2021-12-06
Remove github-workflows job

ssbarnea:
Github no longer prevents merging changes that
alter content of files inside `.github/workflows`.

This change was needed in order to [enable github actions on ansible-navigator](https://dashboard.zuul.ansible.com/t/ansible/buildset/2112fac874e740fb88c32ab87d27a8c3) projects, while still wanting to keep using zuul too.

Depends-On: https://github.com/ansible/ansible-builder/pull/312

PR #1258, opened 2021-12-06
system-required: remove github-workflows

goneri:
We don't need the jobs anymore.

See: https://github.com/ansible/ansible-zuul-jobs/pull/1255


ssbarnea:
typo in title, missing an i

PR #1284, opened 2022-01-10
c.vmware: drop 2.9 and 2.10 sanity tests

goneri:
- drop ansible-test-sanity-docker-stable-2.{9.10}
- don't run the govcsim checks during the gating, the jobs are
  failing and non-voting anyway.

See: https://github.com/ansible-collections/community.vmware/pull/1182


goneri:
recheck

PR #1303, opened 2022-01-17
c.vmware: turn govcsim voting again

goneri:
Depends-On: https://github.com/ansible-collections/community.vmware/pull/1193


goneri:
reload

goneri:
See: https://github.com/ansible/zuul-config/pull/199

PR #1304, opened 2022-01-19
introduce the ansible-ee-tests template

goneri:
Based on network-ee-test with the following differences:

The jobs build the EE container themselves:

- no network-ee-build jobs
- no cache registry in the middle
- simplify the job hierachy. All the job inherite from on single job.
- run on Fedora-35, this can be revisited later.


goneri:
recheck

PR #1312, opened 2022-01-27
sf: move more tox jobs in containers

goneri:
Depends-On: https://github.com/ansible/ansible-zuul-jobs/pull/1313


goneri:
recheck

goneri:
recheck

PR #1325, opened 2022-02-01
cloud: run more jobs in containers

goneri:


goneri:
This does not work as expected with `ansible-test-units-community-aws-python38`.

PR #1357, opened 2022-02-15
{amazon,community}.aws: cross-project testing

goneri:
If there is a change in module_utils, run the CI only for the modules
affected by that change.
This also solves the case where in the same PR there is a change in a
module_utils file (e.g., rds) and a change on a module (e.g., ec2_vpc_igw),
returning the 'target_to_test' containing ec2_vpc_igw and all modules
affected by the change in module_utils. At the moment, 'target_to_test'
is returned with only ec2_vpc_igw.

Co-Authored-by: Alina Buzachis <abuzachis@redhat.com>


goneri:
@alinabuzachis, the PR is ready for review.

PR #1359, opened 2022-02-22
aws test splitter - add lookup plugins

abikouo:


abikouo:
@goneri this aims to add targets for lookup, it was missing from your PR and needed to test the PR https://github.com/ansible-collections/amazon.aws/pull/591
later we should to that for others plugins (filter...)

goneri (approved):
Thank you @abikouo

PR #1373, opened 2022-03-04
kubernetes.core: use fedora-35-1vcpu

goneri:
Use fedora-35-1vcpu because the centos-8-stream is not yet
available. https://softwarefactory-project.io/r/c/config/+/24240
will address the situation. For now, we switch on Fedora 35.


gravesm:
What is the status of this change? We have k8s PRs that depend on this one, but they are failing because of this change from what I can tell.

PR #1388, opened 2022-03-11
DNM - Do not default to all target when none found

abikouo:


goneri:
+1. We need a unit-test for this.

abikouo:
> +1. We need a unit-test for this.

This is not expected to be merged

PR #1389, opened 2022-03-11
migrate eos to ansible-ee model

goneri:
- Migrate from network-ee to ansible-ee
- Group the httpapi and network_cli targets together for simplicity, this can be revisited later.


zuul.d/arista-eos-jobs.yaml
```diff
@@ -49,117 +30,74 @@
...
-    name: network-ee-integration-tests-arista-eos-httpapi-stable-2.11
-    parent: network-ee-integration-tests-arista-eos-stable-2.11
-    vars:
-      test_ansible_skip_tags: network_cli
+      ansible_runner_container_version: stable-2.12-devel
 
 - job:
-    name: network-ee-integration-tests-arista-eos-libssh-stable-2.11
-    parent: network-ee-integration-tests-arista-eos-stable-2.11
+    name: ansible-ee-integration-arista-eos-httpapi-libssh-latest
```
GomathiselviS, line 74: @goneri libssh is not a httpapi connection. We need to remove httpapi from the job name.

PR #1390, opened 2022-03-14
ansible-test-splitter updates

abikouo:


zuul.d/project-templates.yaml
```diff
@@ -70,6 +70,7 @@
         - ansible-test-splitter:
             required-projects:
               - name: github.com/ansible-collections/community.aws
+              - name: github.com/ansible-collections/amazon.aws
```
goneri, line 73: I'm not sure we need this one since the job is already associated with this project. Anyway, it won't hurt.

PR #1405, opened 2022-03-17
nxos: migration to the ansible-ee model

goneri:
Prepare the migration the SoftwareFactory environment.


roles/ansible-test-inventory/templates/inventory-nxos.j2
```diff
@@ -19,8 +19,8 @@ imagemr="1"
 imagetag="I5"
 image_version="7.0(3)I5(1)"
 ipv6_supported=true
-major_version="7.0"
```
goneri, line 22: @GomathiselviS I'm not sure what do to here.

goneri:
recheck

PR #1409, opened 2022-03-18
ansible-test-units: don't generate the coverage report

goneri:
This feature depends on the coverage module that may not be available.
As far as I know, nobody really actually care about the coverage report
anyway.


goneri:
recheck

alinabuzachis:
Can we merge this?

PR #1422, opened 2022-03-25
Revert "use regular VM until the k1s problem is resolved"

goneri:
This reverts commit 05d7f8cd9765881960ed9a7f6a0e237c7c625a61.


goneri:
recheck

PR #1424, opened 2022-03-28
Fix ansible-test-splitter for amazon collections

abikouo:
Fix ansible-test-splitter for amazon collections


abikouo:
@goneri could you please merge this ?

PR #1425, opened 2022-03-28
Add python 3.9 jobs to eos

GomathiselviS:


zuul.sf.d/nodesets.yaml
```diff
@@ -66,6 +66,18 @@
         nodes:
           - eos-4.24.6
 
+- nodeset:
+    name: eos-4.24.6-python39
```
goneri, line 70: Just keep one nodeset please. BTW, both Fedora-35 and CentOS8 come with the Python versions we need.

goneri (changes requested):
cf comments

trishnaguha:
@GomathiselviS We are good to drop Python 2.7

zuul.d/jobs.yaml
```diff
@@ -270,6 +270,14 @@
     vars:
       ansible_test_python: 3.8
 
+- job:
```
NilashishC, line 273: Should we merge a PR with just this change, before we add the 3.9 unit jobs for the other platforms?

goneri (approved):
:+1:

PR #1441, opened 2022-03-31
Update inventory for NX-OS

NilashishC:


roles/ansible-test-inventory/templates/inventory-nxos.j2
```diff
@@ -7,21 +7,22 @@
 nxos-9 ansible_host={{ hostvars['nxos'].ansible_host }} ansible_user=zuul
 
 [nxos:vars]
-ansible_connection=network_cli
+ansible_connection={% if ansible_test_collections is defined -%}ansible.netcommon.{%- endif %}network_cli
+ansible_network_os={% if ansible_test_collections is defined -%}cisco.nxos.{%- endif %}nxos
```
NilashishC, line 11: This makes sure that the cliconf from the collection is pulled and not ansible/ansible one when run with stable-2.9.

PR #1448, opened 2022-04-04
Add ansible-ee nxos test jobs

GomathiselviS:

### DIFF
--- a/zuul.d/cisco-nxos-jobs.yaml
+++ b/zuul.d/cisco-nxos-jobs.yaml
@@ -1,40 +1,32 @@
 ---
-# ansible-collections/cisco.nxos jobs
-
+# ansible-ee
 - job:
-    name: ansible-network-nxos-appliance
-    parent: ansible-network-appliance-base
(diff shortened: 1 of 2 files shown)


zuul.d/cisco-nxos-jobs.yaml
```diff
@@ -1,40 +1,32 @@
...
-      - name: github.com/ansible/ansible
-      - name: github.com/ansible-collections/cisco.nxos
     timeout: 10800
-    vars:
-      ansible_collections_repo: github.com/ansible-collections/cisco.nxos
-      ansible_test_command: network-integration
-      ansible_test_integration_targets: "nxos_.*"
+    files:
+      - ^plugins/.*$
+      - ^tests/integration/.*$
```
NilashishC, line 29: We should also include requirements.txt to this, I believe. This would make sure that PRs like this https://github.com/ansible-collections/cisco.nxos/pull/447 would also be tested.

goneri: Indeed, nice catch, we can also add test-requirements.txt.

PR #1451, opened 2022-04-05
Add Py 3.9 jobs for IOS removing Py 2.7

KB-perByte:


trishnaguha:
For stable-2.9 Python 3.9 won't work. We should restrict 3.8 for 2.9 branch https://github.com/ansible/ansible-zuul-jobs/pull/1455

zuul.d/cisco-ios-jobs.yaml
```diff
@@ -98,13 +103,15 @@
...
     vars:
+      test_ansible_python_interpreter: /usr/bin/python3.8
       ansible_runner_container_version: stable-2.9-devel
 
 # - network-ee-integration-tests-ios-libssh-stable-2.11
 - job:
     name: ansible-ee-integration-ios-libssh-stable-2.11
     parent: ansible-ee-integration-ios-libssh-latest
     vars:
+      test_ansible_python_interpreter: /usr/bin/python3.8
```
trishnaguha, line 114: Stable-2.11 should not use Python 3.8 interpreter.

PR #1458, opened 2022-04-06
Restricting python version to 3.8 for ansible-2.9 jobs

rohitthakur2590:


zuul.ansible.d/nodesets.yaml
```diff
@@ -575,6 +575,21 @@
         nodes:
           - fedora-35
 
+- nodeset:
+    name: vsrx3-18.4R1-python39
```
GomathiselviS, line 579: We have decided to maintain a generic name for the nodes and not to use the python versions along with it. Can you please remove pyhton39 from the name and modify the references?

goneri (approved):

goneri:
regate

goneri (approved):

PR #1472, opened 2022-04-07
Set amazon.cloud integration target

gravesm:
This sets the integration target for the amazon.cloud collection so that all the tests will be run.

### DIFF
--- a/zuul.d/ansible-cloud-jobs.yaml
+++ b/zuul.d/ansible-cloud-jobs.yaml
@@ -382,7 +382,7 @@
     vars:
       ansible_test_command: integration
       ansible_test_python: 3.9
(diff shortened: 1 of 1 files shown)


zuul.d/ansible-cloud-jobs.yaml
```diff
@@ -438,7 +438,7 @@
       ansible_test_collections: true
       ansible_test_command: integration
       ansible_test_python: 3.8
-      ansible_test_integration_targets: ""
+      ansible_test_integration_targets: zuul/aws/cloud_control
```
goneri, line 441: Add a / at the end: `zuul/aws/cloud_control/`

gravesm: Still nothing.

PR #1481, opened 2022-04-11
ansible.yang python38 and python39 support

rohitthakur2590:
Signed-off-by: Rohit Thakur <rohitthakur2590@outlook.com>

### DIFF
--- a/zuul.d/jobs.yaml
+++ b/zuul.d/jobs.yaml
@@ -340,6 +340,8 @@
     parent: ansible-test-units-base-python36
     required-projects:
       - name: github.com/ansible-collections/ansible.yang
+      - name: github.com/ansible/ansible
+        override-checkout: stable-2.9
     vars:
       ansible_collections_repo: github.com/ansible-collections/ansible.yang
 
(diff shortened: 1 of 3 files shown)


zuul.d/junipernetworks-junos-jobs.yaml
```diff
@@ -265,6 +265,9 @@
     name: ansible-test-network-integration-junos-vsrx-python36
     parent: ansible-test-network-integration-junos-vsrx
     nodeset: vsrx3
+    required-projects:
+      - name: github.com/ansible/ansible
+        override-checkout: stable-2.11
```
GomathiselviS, line 270: why do we use stable-2.11 here? I think it should be devel.

GomathiselviS (changes requested):
- python27 unit test job has to be removed.
- python36 and python37 unit test job has to use ansible stable-2.9 branch.

zuul.d/jobs.yaml
```diff
@@ -367,20 +359,42 @@
...
 
+- job:
+    name: ansible-test-network-integration-ansible-yang-iosxr-netconf-python38
+    parent: ansible-test-network-integration-iosxr-python38
+    vars:
+      ansible_collections_repo: github.com/ansible-collections/ansible.yang
+      ansible_test_integration_targets: ""
+
 - job:
     name: ansible-test-network-integration-ansible-yang-junos-vsrx-netconf-python36
```
GomathiselviS, line 385: @rohitthakur2590 All the network-ee integration jobs for the network collections are tested only with py39. If you want to test this collection with py36, it can tested with ansible-2.9 and not devel, because 'ansible-core' requires a different Python:  '>=3.8'`

PR #1494, opened 2022-04-19
downstream-ee testing for amazon.aws and vmware.vmware_rest

goneri:
Depends-On: https://github.com/ansible-collections/vmware.vmware_rest/pull/318

New set of periodical jobs that run the last snapshot of the Downstream-EE
image against amazon.aws and vmware.vmware_rest.


goneri:
recheck

goneri:
recheck

PR #1498, opened 2022-04-20
Make ansible-ee-tests-latest globally non-voting and ansible-ee-tests-stable-2.12 non voting for amazon.aws

tremble:
ansible-latest (ie devel) is expected to break from time to time, with things like new sanity tests landing.  Such changes will likely result in new ignore file entries being needed until cleanup can be applied.  While we should be reacting reasonably quickly to this, we shouldn't instantly break CI for multiple collections.

ansible-ee-tests-stable-2.12 and ansible-ee-tests-latest are currently broken for amazon.aws due to #1497 this is a problem with the containers, the unit and sanity tests are running fine in the 'standard' environments.

### DIFF
--- a/zuul.d/project-templates.yaml
+++ b/zuul.d/project-templates.yaml
@@ -102,6 +102,10 @@
         - integration-community.aws-11
         - integration-community.aws-12
         - integration-community.aws-13
+        # Failing due to broken six/urllib3 install in generated containers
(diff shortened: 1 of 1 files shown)


tremble:
CC: @alinabuzachis @goneri @GomathiselviS

PR #1509, opened 2022-04-27
ansible-test-sanity-* don't pin on py38

goneri:


GomathiselviS (approved):

gundalow:
Does this remove `--requirements` as well? That's not needed when using `--docker`

goneri:
> Does this remove `--requirements` as well? That's not needed when using `--docker`

This was done yesterday with https://github.com/ansible/ansible-zuul-jobs/pull/1508

PR #1510, opened 2022-04-28
downstream-ee/aws: call runme.sh from within the container

goneri:
Depends-On: https://github.com/ansible-collections/amazon.aws/pull/797


goneri:
recheck

goneri:
regate

PR #1523, opened 2022-05-05
Remove pinned k8s.core version for okd

gravesm:
I don't remember why this was here, and I think we should probably just
remove it. My guess is this may be the reason why our sanity tests have
been intermittently failing, lately.


gravesm:
recheck

PR #1526, opened 2022-05-10
ansible-test-sanity-docker-milestone non-voting

goneri:
There was a bug introduced in the latest milestone that broke k8s.core.
We hate to just keep making ci jobs non-voting, but we also don't want
unreleased ansible code to block PRs

See: https://github.com/ansible/ansible/issues/77764


gravesm:
Thanks! I wish there were a better solution for these kinds of problems, but I don't know what it would look like.

PR #1533, opened 2022-05-12
okd: disable sanity tests with ansible-2.9

goneri:
As decided on Slack.


gravesm:
Let's actually keep the upstream as that does not have issues. This would also likely catch any problems for the downstream.

PR #1534, opened 2022-05-13
add 2.13 sanity jobs to vmware_rest and cloud.common

goneri:


mariolenz:
@Gomathiselvi-S This PR is quite old, and with your merged PR #1838 I think it doesn't make much sense anymore. Shouldn't it be closed?

PR #1539, opened 2022-05-16
Fix OKD downstream sanity tests

gravesm:
The downstream collection bulid process needs the yaml package. I don't
know why, but this recently started failing saying it couldn't find
yaml. This also makes the milestone sanity tests non-voting as there is
currently a bug in the milestone branch that leads to failures.


goneri:
I'm working on a better fix that tackle the original problem.

PR #1541, opened 2022-05-16
Remove ansible.utils jobs entirely

cidrblock:
- This leaves ansible.utils as a dep for other projects, but removes it's jobs.


cidrblock:
recheck

PR #1553, opened 2022-05-18
To fix py39->py36 for TM deepsec unit test case

justjais:
To fix py39->py36 for TM deepsec unit test case


justjais:
cc. @goneri plz review and merge

PR #1584, opened 2022-07-05
Remove IOS unit and sanity

KB-perByte:
Depends-On: https://github.com/ansible/zuul-config/pull/416

Remove IOS unit and sanity as implemented with GHA


goneri:
You also need to remove all the references from zuul.d/projects.yaml, c.f. the error.

PR #1585, opened 2022-07-05
Remove IOSXR unit and sanity jobs

KB-perByte:
Depends-On: https://github.com/ansible/zuul-config/pull/416

Remove IOSXR unit and sanity jobs as they are addressed in GHA


goneri:
recheck

PR #1595, opened 2022-07-22
vcenter: deploy and use an external identity provider

goneri:
Depends-On: https://github.com/ansible/ansible-zuul-jobs/pull/1600

Deploy an OpenLDAP instance and use it as external identity source.

Accounts:

- lisa@datastore.test / useruser
- pierre@datastore.test / useruser
- george@datastore.test / useruser - admin user


goneri:
Not needed anymore.

PR #1599, opened 2022-07-25
Bump sanity tests to Python 3.9 - devel branch has dropped support for Python 3.8

tremble:
Devel tests are currently failing because (as announced) Ansible core devel branch has now dropped support for Python 3.8.

For example: https://ansible.softwarefactory-project.io/zuul/build/cbdb49248070403589d008f339344c48

### DIFF
--- a/zuul.d/ansible-test-jobs.yaml
+++ b/zuul.d/ansible-test-jobs.yaml
@@ -2,12 +2,12 @@
 - job:
     name: ansible-test-sanity-docker
     parent: unittests
-    nodeset: controller-python38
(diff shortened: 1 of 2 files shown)


goneri:
I was about to push the following PR to address the same problem:

- https://github.com/ansible/ansible-zuul-jobs/pull/1597
- https://github.com/ansible/ansible-zuul-jobs/pull/1602

goneri:
Thank @tremble for the PR. The two PR I mentioned above should resolve the problem.

Issue #1598, opened 2022-07-25
Some ansible-ee-integration-arista-eos-libssh jobs missing nslookup binary


sc68cal:
❌ [ansible-ee-integration-arista-eos-libssh-latest ](https://ansible.softwarefactory-project.io/zuul/build/9baeeac2b75a4c52ae209084bc1018ae)RETRY_LIMIT in 12m 16s
❌ [ansible-ee-integration-arista-eos-libssh-stable-2.11 ](https://ansible.softwarefactory-project.io/zuul/build/493a13568de94365b3c622ed44ebfb6a)RETRY_LIMIT in 13m 00s
❌ [ansible-ee-integration-arista-eos-libssh-stable-2.12 ](https://ansible.softwarefactory-project.io/zuul/build/24e9e72691fe47148d6dabef9906a24a)RETRY_LIMIT in 12m 22s

PR #1601, opened 2022-07-25
enable-fips now requires nslookup

goneri:
See: https://opendev.org/zuul/zuul-jobs/src/branch/master/roles/post-reboot-tasks


sc68cal:
This fixes #1598

sc68cal (approved):

PR #1603, opened 2022-07-25
c.vmware: test ansible 2.12, ignore py<3.8

goneri:
- disable unit-tests for Python versions older than 3.8
- use the current Python version of the controller (3.10 on Fedora 36)
  for the integration tests
- remove the Python version from the job name, since we don't really
  care and the name is long enough.

### DIFF
--- a/zuul.d/ansible-cloud-jobs.yaml
+++ b/zuul.d/ansible-cloud-jobs.yaml
@@ -14,7 +14,7 @@
         ansible_network_os: vmware_rest
       esxi1:
         ansible_network_os: vmware_rest
-    nodeset: vmware-vcsa-7.0.3-python36
+    nodeset: vmware-vcsa-7.0.3
 
 
 # master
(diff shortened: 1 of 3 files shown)


zuul.d/ansible-cloud-jobs.yaml
```diff
@@ -79,34 +76,33 @@
...
       ansible_test_do_number: 1
 
 - job:
-    name: ansible-test-cloud-integration-vcenter7_1esxi-python36-stable211_2_of_2
-    parent: ansible-test-cloud-integration-vcenter7_1esxi-python36
+    name: ansible-test-cloud-integration-vcenter7_1esxi-stable212_2_of_2
+    parent: ansible-test-cloud-integration-vcenter7_1esxi
     required-projects:
       - name: github.com/ansible/ansible
         override-checkout: stable-2.11
```
mariolenz, line 108:
```suggestion
        override-checkout: stable-2.12
```

zuul.d/ansible-cloud-jobs.yaml
```diff
@@ -116,8 +112,8 @@
 
 
 - job:
-    name: ansible-test-cloud-integration-vcenter7_2esxi-python36-stable211
-    parent: ansible-test-cloud-integration-vcenter7_2esxi-python36
+    name: ansible-test-cloud-integration-vcenter7_2esxi-stable212
+    parent: ansible-test-cloud-integration-vcenter7_2esxi
     required-projects:
       - name: github.com/ansible/ansible
         override-checkout: stable-2.11
```
mariolenz, line 119:
```suggestion
        override-checkout: stable-2.12
```

mariolenz (commented):
Since you've changed the name from stable211 to stable212, shouldn't the `override-checkout` been changed, too?

zuul.d/ansible-cloud-jobs.yaml
goneri: Thanks!

zuul.d/ansible-cloud-jobs.yaml
goneri: +1

mariolenz (approved):
LGTM!

PR #1604, opened 2022-07-25
vmware_rest: update the tests

goneri:
- integration: use the current Python version on the controller
- add Ansible-2.13 sanity tests


goneri:
recheck

goneri:
recheck

PR #1608, opened 2022-08-04
Add amazon.aws as a dependency of community.aws in third-party checks

tremble:
Without building the dependency we can't properly set up the dependencies for the main branch which still contains unreleased code.


tremble:
@goneri

tremble:
Thanks @goneri

PR #1617, opened 2022-08-29
Allow 'trivial' changelog fragments

tremble:
There are times when there's no need for a 'public' changelog isn't necessary (eg test-only changes, or module-docs work)

Based on comments from @felixfontein the appropriate way to deal with this is to allow 'trivial' fragments:

`always remember the 'trivial:' category in changelog fragments, it's exactly for such cases where you don't want to clutter the changelog with such things, but have to write a fragment`


tremble:
@goneri

PR #1625, opened 2022-09-13
list_changed_targets: drop build_up_batches

goneri:
Use `build_up_batches_by_time()` by default. By default, we assume a target
takes 180s.


abikouo (approved):
Nice, Thanks

goneri:
> Nice, Thanks

Merci Aubin :-).

PR #1634, opened 2022-09-23
ansible-test-splitter: share equally targets accross jobs

abikouo:
Update `ansible-test-splitter` in order to use all available slots instead of putting under pressure some jobs


goneri:
I've mixed feeling about this change, meanly because of AWS. Each time we start a new job, we've got a 5 minutes overhead to prepare the instance, fetch the collections, install the dependencies, etc. In addition, there is a semaphore that prevent the run of more than 8 jobs in parallel.

I've the feeling we can reach a same behaviour if we properly set the `time=XXm` entries in the `aliases` file of the slowest targets.

jillr (approved):
We're seeing timeouts with multiple collections without this change, let's try merging and see how it goes for a while - we can revert if it has an overall negative impact.

PR #1639, opened 2022-09-26
add ansible-test-sanity-docker-stable-2.14

goneri:


felixfontein:
More info on that sanity problem is here: https://github.com/ansible/ansible/issues/78882#issuecomment-1261219764

PR #1654, opened 2022-10-05
use amazon.aws unit test as the template for community.aws unit tests instead of community.vmware

tremble:
community.vmware is hard coded to look for community.vmware.  This broke community.aws CI


tremble:
CC @goneri 

See for example:
https://ansible.softwarefactory-project.io/zuul/buildset/7378cca9cd3a45df9470812deb973029

PR #1662, opened 2022-10-20
add the possibility to define extra tests for ansible-test-splitter

abikouo:
For the collection using the `ansible-test-splitter` role, users can add the following line in the pull request description

#### Define list of ansible release to test
Zuul-Releases: 2.9 devel
#### Define list of extra targets to include to ansible-test
Zuul-Extra-Targets: target1 target2 target3
#### Define list of targets for ansible-test
Zuul-Targets: target1

### DIFF
--- a/roles/ansible-test-splitter/files/list_changed_targets.py
+++ b/roles/ansible-test-splitter/files/list_changed_targets.py
@@ -52,7 +52,6 @@ parser.add_argument(
 parser.add_argument(
     "--pull-request",
     dest="pull_request",
-    required=True,
     type=int,
     help="GitHub Pull request number. e.g: 51",
 )
@@ -60,7 +59,6 @@ parser.add_argument(
 parser.add_argument(
     "--project-name",
     dest="project_name",
-    required=True,
     type=str,
     help="GitHub project name. e.g: ansible-collections/kubernetes.core",
 )
@@ -467,7 +465,11 @@ if __name__ == "__main__":
 
     ansible_releases = args.ansible_releases
     zuul_targets, zuul_extra_targets = [], []
-    pr_request = read_pullrequest_zuul_override(args.project_name, args.pull_request)
+    pr_request = None
+    if args.project_name is not None and args.pull_request is not None:
+        pr_request = read_pullrequest_zuul_override(
+            args.project_name, args.pull_request
+        )
 
     if pr_request:
         release_to_test = pr_request.get(ZUUL_RELEASES)
--- a/roles/ansible-test-splitter/files/list_changed_targets.py
+++ b/roles/ansible-test-splitter/files/list_changed_targets.py
@@ -59,11 +59,19 @@ parser.add_argument(
 )
 
 parser.add_argument(
-    "--change-url",
-    dest="change_url",
+    "--pull-request",
+    dest="pull_request",
+    required=True,
+    type=int,
+    help="GitHub Pull request number. e.g: 51",
+)
+
+parser.add_argument(
+    "--project-name",
+    dest="project_name",
     required=True,
     type=str,
-    help="GitHub Pull request URL. e.g: https://api.github.com/repos/ansible-collections/kubernetes.core/pulls/51",
+    help="GitHub project name. e.g: ansible-collections/kubernetes.core",
 )
 
 parser.add_argument(
@@ -430,15 +438,19 @@ class ElGrandeSeparator:
         return result
 
 
-def read_pullrequest_body(change_url):
+def read_pullrequest_body(project_name, pull_request):
     if REQUESTS_MODULE_IMPORT_ERROR:
         raise REQUESTS_MODULE_IMPORT_ERROR
+    change_url = "https://api.github.com/repos/%s/pulls/%d" % (
+        project_name,
+        pull_request,
+    )
     return [x for x in requests.get(change_url).json().get("body").split("\n") if x]
 
 
-def read_user_extra_requests(change_url):
+def read_user_extra_requests(project_name, pull_request):
 
-    desc = read_pullrequest_body(change_url)
+    desc = read_pullrequest_body(project_name, pull_request)
     extra_requests_patterns = (
         "Zuul-Test-Include-Extra-Targets",
         "Zuul-Test-with-Targets",
@@ -469,7 +481,7 @@ if __name__ == "__main__":
 
     ansible_releases = args.ansible_releases
     zuul_targets, zuul_extra_targets = [], []
-    pr_request = read_user_extra_requests(args.change_url)
+    pr_request = read_user_extra_requests(args.project_name, args.pull_request)
 
     if pr_request:
         release_to_test = pr_request.get("Zuul-Test-with-Releases")
--- a/roles/ansible-test-splitter/files/list_changed_targets.py
+++ b/roles/ansible-test-splitter/files/list_changed_targets.py
@@ -11,12 +11,15 @@ import re
 from collections import defaultdict
 
 try:
-    from github import Github
 
-    HAS_GITHUB_MODULE = True
+    import requests
 
-except ImportError:
-    HAS_GITHUB_MODULE = True
+    REQUESTS_MODULE_IMPORT_ERROR = None
+
+except ImportError as e:
+
+    REQUESTS_MODULE_IMPORT_ERROR = e
+
 
 parser = argparse.ArgumentParser(
     description="Evaluate which targets need to be tested."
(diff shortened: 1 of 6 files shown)


roles/ansible-test-splitter/files/list_changed_targets.py
```diff
@@ -10,6 +10,13 @@
 import re
 from collections import defaultdict
 
+try:
```
goneri, line 13: You can do the same with just `requests` or `urllib`. I would prefer we avoid third party dependency in the CI. Also, I don't think we need this try/except block. It's better if the job fails immediately. This will simplify the troubleshooting.

roles/ansible-test-splitter/files/list_changed_targets.py
```diff
@@ -412,14 +435,69 @@ def build_result_struct(self, batches):
...
+            return [x for x in pr.body.split("\n") if x]
+
+    except:
+        return []
+
+
+def read_user_extra_requests(project_name, pr_number):
+
+    desc = read_pullrequest_body(project_name, pr_number)
+    ansible_test_regex = (
```
goneri, line 454: Those are not regex, I would call them `extra_requests_patterns` instead to avoid the confusion.

roles/ansible-test-splitter/tasks/main.yaml
```diff
@@ -1,31 +1,60 @@
...
-  zuul_return:
-    data: "{{ for_zuul_return.data }}"
+- block:
+    - name: Create temporary directory for tests
+      tempfile:
+        state: directory
+        suffix: .splitter
+      register: _result
+
+    - set_fact:
```
goneri, line 9: I don't think we need this, we've got already the `json` and `requests` packages in the container that we use to run the job:

```
⋊> ~ podman run quay.io/software-factory/zuul-worker-ansible python -c 'import json;print(json.__version__)'                                                                                                                          10:48:12
2.0.9
⋊> ~ podman run quay.io/software-factory/zuul-worker-ansible python -c 'import requests;print(requests.__version__)'                                                                                                                  10:48:19
2.27.1
```

goneri (changes requested):

gravesm (commented):
I like the idea behind this, but I have a few comments/suggestions. The PR keywords are kind of lengthy and it seems like the sort of thing we'd have to lookup every time. I would suggest something shorter like: `zuul-targets` and `zuul-releases`. I would also suggest doing a case insensitive comparison.

What's the difference between test-with-targets and test-include-extra-targets?

roles/ansible-test-splitter/files/list_changed_targets.py
```diff
@@ -10,6 +10,16 @@
 import re
 from collections import defaultdict
 
+try:
+
+    import requests
```
goneri, line 15: I don't think we need this kind of construction. It's better to fail fast if the environment is not as expected.

roles/ansible-test-splitter/files/list_changed_targets.py
```diff
@@ -412,14 +438,65 @@ def build_result_struct(self, batches):
...
+        data = (":".join(line.split(":")[1:])).replace(",", " ")
+        return [x for x in data.split() if x]
+
+    result = {}
+    for line in desc:
+        try:
+            for key in extra_requests_patterns:
+                if line.startswith(key + ":"):
+                    result[key] = _extract_data(line)
+        except:
```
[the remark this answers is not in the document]
goneri: oh there is another catch-all block in `build_import_tree()`.

abikouo:
> I like the idea behind this, but I have a few comments/suggestions. The PR keywords are kind of lengthy and it seems like the sort of thing we'd have to lookup every time. I would suggest something shorter like: `zuul-targets` and `zuul-releases`. I would also suggest doing a case insensitive comparison.
> 
> What's the difference between test-with-targets and test-include-extra-targets?

ok thanks for the suggestion, I will simplify the keyword names and add case insensitive comparison.
here is the difference between `zuul-targets` and `zuul-extra-targets`

Assume I have a PR that is supposed to test `case` and `case2`. When I add `zuul-targets: case3` my PR will test only `case3` but if it is `zuul-extra-targets: case3` it will test `case1`, `case2` and `case3`

roles/ansible-test-splitter/files/list_changed_targets.py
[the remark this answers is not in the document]
abikouo: I can remove this one

roles/ansible-test-splitter/files/list_changed_targets.py
abikouo: The code will fail a bit later with the same behavior.
I did this to avoid adding `requests` as requirement for unit testing knowing that this is not used and could potentially been replaced by another library with the same behavior

goneri: The unit-tests won't run in the same environment, it's a different thing. We've got the controller on the container that we use to run `ansible-test-splitter`. Right now, it's basically a Fedora with the ansible-core package.

If we were to upgrade the container with something that doesn't include requests, we would prefer a quick failure of the job. This instead of discovering 1 month later that in some specific cases the job fail. It's often better to fail fast.

abikouo: I remove the `try...except` block, unit tests are failing now

roles/ansible-test-splitter/files/list_changed_targets.py
```diff
@@ -48,6 +49,22 @@
     help="ansible release version to test for each jobs",
 )
 
+parser.add_argument(
+    "--pull-request",
+    dest="pull_request",
+    required=True,
```
goneri, line 55: Could we have `required=False` by default. This will simplify the troubleshooting when we run the command locally.

roles/ansible-test-splitter/files/list_changed_targets.py
```diff
@@ -48,6 +49,22 @@
...
+    type=int,
+    help="GitHub Pull request number. e.g: 51",
+)
+
+parser.add_argument(
+    "--project-name",
+    dest="project_name",
+    required=True,
```
goneri, line 63: The same.

roles/ansible-test-splitter/files/list_changed_targets.py
```diff
@@ -412,14 +429,60 @@ def build_result_struct(self, batches):
...
 
     collections = [Collection(i) for i in args.collection_to_tests]
     collections_names = [c.collection_name() for c in collections]
 
+    ansible_releases = args.ansible_releases
+    zuul_targets, zuul_extra_targets = [], []
+    pr_request = read_pullrequest_zuul_override(args.project_name, args.pull_request)
+
+    if pr_request:
+        release_to_test = pr_request.get(ZUUL_RELEASES)
```
goneri, line 473: I believe the `zuul-releases` is not used anymore. Can we remove it?

roles/ansible-test-splitter/files/list_changed_targets.py
abikouo: ok

roles/ansible-test-splitter/files/list_changed_targets.py
abikouo: It is

roles/ansible-test-splitter/files/list_changed_targets.py
```diff
@@ -412,14 +427,64 @@ def build_result_struct(self, batches):
...
+    if args.project_name is not None and args.pull_request is not None:
+        pr_request = read_pullrequest_zuul_override(
+            args.project_name, args.pull_request
+        )
+
+    if pr_request:
+        release_to_test = pr_request.get(ZUUL_RELEASES)
+        if release_to_test:
+            tmp_releases = [rel for rel in release_to_test if rel in ansible_releases]
+            ansible_releases = tmp_releases
```
gravesm, line 478: I'm not sure I understand the use case behind this. What's the case where we'd want to limit the releases tested against on a single PR? This seems like the sort of thing that should be addressed in zuul config, not as a one-off on a PR. If a PR is failing for a specific release and we just ignore that release, then it's still going to be failing for every other PR.

abikouo: The scope of this change is debugging only, we are not supposed to merge any PR using these keywords
In case you want to limit the number of ansible releases to be tested for any reason, you can use this

goneri: This bypass the `override-checkout` feature that is already in place. If you want to test a PR against a different branch, you can still prepare a quick PR in ansible-zuul-jobs that adjust the override-checkout, and Depends-On it. I don't think we need something like that in this script.

abikouo: all the keywords here are bypassing the `override-checkout` here so this PR makes no sense, it is a burden pain to create what you call a 'quick PR' each time you need to perform a particular test
This PR is not really for my debug, if you think it is not useful I will close it, but for me I found the utility of this

goneri (approved):

PR #1667, opened 2022-10-30
Bump splitter total_job for aws collections

tremble:
amazon.aws PRs which change module_utils can trigger almost all modules.

Having the slow amazon.aws jobs split up has resulted in us going beyond 18 total jobs.


tremble:
@goneri

goneri:
Thank you Mark!

PR #1668, opened 2022-11-03
Skip all 'indirect' targets if they're in an ignored category

tremble:
When making changes to module_utils we can trigger large numbers of integration tests.

If we wouldn't run the tests during a "test all" scenario we shouldn't trigger the tests just because they depend on something in module_utils


tremble:
CC: @goneri

goneri (approved):

PR #1686, opened 2022-11-07
CI TESTING

goneri:


mariolenz:
@goneri Since this looks like just a test, and an old one, could you please close it? I don't think it makes sense to keep it open.

Additionally, there are conflicts with the current master branch.

PR #1687, opened 2022-11-08
Update AWS lookup plugin test names

tremble:
SSIA


goneri:
The file is autogenerated.

goneri (approved):

PR #1710, opened 2022-12-01
Add integration test job to cloud.terraform

gravesm:


playbooks/ansible-cloud/terraform/pre.yaml
```diff
@@ -0,0 +1,19 @@
+---
+- hosts: all
+  tasks:
+    - name: Get current Terraform version
+      uri:
+        url: https://checkpoint-api.hashicorp.com/v1/check/terraform
```
goneri, line 6: There is also a RPM repository: https://developer.hashicorp.com/terraform/cli/install/yum it may easier to follow this path.

gravesm: Ah, yeah, thanks.

gravesm:
Closing in favor of #1749

PR #1730, opened 2022-12-22
vmware.vmware_rest: migrate to gouttelette

goneri:


mariolenz:
I'm not 100% sure but I think #1892 makes this PR obsolete. Should it be closed? :thinking:

PR #1731, opened 2023-01-02
Use fake SELinux bindings for Python 3.9 on Fedora

tremble:
F35 bumped the default Python3 to 3.10, this in turn means that python3-libselinux only supports Python 3.10, and something pulling `selinux` into the venv dependencies causes trouble

For example:
https://abbbf6c58851451ba48a-b8cc0d14874e33ae481664022b4871ba.ssl.cf5.rackcdn.com/1633/c09a47bdf253cf46dd290ac83c1cbef54d4d1635/check/ansible-test-units-amazon-aws-python39/b47386c/job-output.txt
```
ERROR! Unexpected Exception, this is probably a bug: Failed to detect selinux python bindings
```

and https://7ef6fc4cf02dfb7892e0-9d66b51498e3b64d406ee8f92531e474.ssl.cf2.rackcdn.com/1303/dbba2b0add3df9ffba7fcafd3b2c65dc50a269de/gate/integration-amazon.aws-1/0f178e3/job-output.txt
```
Exception: Failed to detect selinux python bindings at
```

Also adds Python 3.10 unit tests for amazon.aws and community.aws since we just `disabled` the selinux module


zuul.d/jobs.yaml
```diff
@@ -196,14 +196,6 @@
...
       ansible_test_python: 3.9
 
-- job:
-    name: ansible-test-units-base-python39
-    parent: ansible-test-units-base
-    nodeset: fedora-35-1vcpu
-    abstract: true
-    vars:
-      ansible_test_python: 3.9
-
```
tremble, line 206: This was a duplicate, overridden by an entry below

tremble:
CC: @goneri

goneri (approved):

PR #1743, opened 2023-01-13
Add support for connection plugin integration tests

tremble:
Changes to connection plugins are currently not triggering integration tests

### DIFF
--- a/roles/ansible-test-splitter/files/list_changed_targets.py
+++ b/roles/ansible-test-splitter/files/list_changed_targets.py
@@ -199,38 +199,52 @@ class WhatHaveChanged:
             ]
         return self.files
 
-    def modules(self):
-        """List the modules impacted by the change"""
-        for d in self.changed_files():
-            if str(d).startswith("plugins/modules/"):
-                yield PosixPath(d)
-
-    def inventory(self):
-        """List the inventory plugins impacted by the change"""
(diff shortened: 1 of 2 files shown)


tremble:
@abikouo could I get a review please

roles/ansible-test-splitter/files/list_changed_targets.py
```diff
@@ -199,38 +199,52 @@ def changed_files(self):
...
-        """List the Python modules impacted by the change"""
+    def lookup(self):
+        """List the lookup plugins impacted by the change"""
+        yield from self._path_matches("plugins/lookup/")
+
+    def modules(self):
+        """List the modules impacted by the change"""
+        yield from self._path_matches("plugins/modules/")
+
+    def _util_matches(self, base_path, import_path):
```
abikouo, line 231: could be simplified using only parameter `import_path` as the util is always part of `plugins`

abikouo (approved):

roles/ansible-test-splitter/files/list_changed_targets.py
tremble: The difference is that it also returns the ansible_collections.ABC.DEF.plugins.module_utils.XZY, which is then used to trigger tests against the dependencies of the utils.

PR #1752, opened 2023-01-23
Ensure amazon.aws and community.aws build dependencies are consistent

tremble:
Ensure amazon.aws and community.aws build dependencies are consistent between Check, On-Demand and Gate

We don't want check passing when gate's still going to fail


tremble:
@abikouo Could I get another review/merge please

abikouo (approved):

PR #1786, opened 2023-04-18
Remove amazon.aws sanity and unit jobs

GomathiselviS:
amazon.aws unit and sanity jobs moved to GHA


alinabuzachis:
cc @gravesm 

Do we need to keep `cloud-tox-py3` on zuul?

gravesm (approved):

alinabuzachis (approved):

PR #1791, opened 2023-04-24
Remove Pravic jobs from zuul

GomathiselviS:
Pravic testing is moved to GHA. Hence removing it from Zuul.


GomathiselviS:
> Oh, I think we would have needed to make this change, first. We may need to re-add the pravic project temporarily to zuul-config in order to get this merged.

We need to merge [zuul-config PR](https://github.com/ansible/zuul-config/pull/507) first , followed by this one.

PR #1797, opened 2023-04-27
Remove amazon.cloud job template

GomathiselviS:
The testing for this collection is moved to GitHub Actions. Hence removing the template from zuul project-config.

### DIFF
--- a/zuul.d/project-templates.yaml
+++ b/zuul.d/project-templates.yaml
@@ -85,26 +85,6 @@
       jobs: *ansible-collections-community-aws-jobs
 
 - project-template:
-    name: ansible-collections-amazon-cloud
-    check:
-      jobs: &ansible-collections-amazon-cloud-jobs
-        - ansible-test-sanity-docker-devel
-        - ansible-test-sanity-docker-milestone
-        - ansible-test-sanity-docker-stable-2.11
-        - ansible-test-sanity-docker-stable-2.12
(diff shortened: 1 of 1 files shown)


gravesm (approved):

zuul.d/project-templates.yaml
```diff
@@ -84,26 +84,6 @@
...
-        - ansible-test-sanity-docker-stable-2.14
-        - build-ansible-collection:
-            required-projects:
-              - name: github.com/ansible-collections/amazon.aws
-              - name: github.com/ansible-collections/community.aws
-              - name: github.com/ansible-collections/community.crypto
-    gate:
-      jobs: *ansible-collections-amazon-cloud-jobs
-    ondemand:
-      jobs: *ansible-collections-amazon-cloud-jobs
```
alinabuzachis, line 105: Not sure though, but we should probably keep these. If not, just ignore my comment.

@gravesm any thoughts?

gravesm: I think these are all covered now by GHA, right @GomathiselviS? Or am I misunderstanding things?

GomathiselviS: I don't know why the required-projects key is used here. The sanity tests do not need these projects. Please correct me if I am wrong.

The GitHub action workflows for unit and sanity pass without these projects.

GomathiselviS:
recheck

PR #1832, opened 2023-11-09
vmware_rest: Remove sanity tests

mariolenz:
Drop outdated ansible (core) versions. Also milestone and devel sanity tests.

### DIFF
--- a/zuul.d/project-templates.yaml
+++ b/zuul.d/project-templates.yaml
@@ -455,11 +455,6 @@
               - name: github.com/ansible-collections/community.vmware
         - ansible-test-sanity-docker-devel
         - ansible-test-sanity-docker-milestone
-        - ansible-test-sanity-docker-stable-2.9
-        - ansible-test-sanity-docker-stable-2.10:
-            voting: false
-        - ansible-test-sanity-docker-stable-2.11
-        - ansible-test-sanity-docker-stable-2.12
         - tox-cloud-refresh-examples-vmware
     gate:
       jobs: *ansible-collections-community-vmware-rest-jobs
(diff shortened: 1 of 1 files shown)


mariolenz:
I think the sanity tests against devel are not really helpful. So let's remove this job also while we're at it.

@Gomathiselvi-S What do you think? Can you approve and merge?

zuul.d/project-templates.yaml
```diff
@@ -453,24 +453,13 @@
...
               - name: github.com/ansible-collections/cloud.common
               - name: github.com/ansible-collections/community.vmware
-        - ansible-test-sanity-docker-devel
         - ansible-test-sanity-docker-milestone
-        - ansible-test-sanity-docker-stable-2.9
-        - ansible-test-sanity-docker-stable-2.10:
-            voting: false
-        - ansible-test-sanity-docker-stable-2.11
-        - ansible-test-sanity-docker-stable-2.12
         - tox-cloud-refresh-examples-vmware
```
GomathiselviS, line 457: We can remove this. It is no more valid.

zuul.d/project-templates.yaml
```diff
@@ -453,24 +453,13 @@
             required-projects:
               - name: github.com/ansible-collections/cloud.common
               - name: github.com/ansible-collections/community.vmware
-        - ansible-test-sanity-docker-devel
         - ansible-test-sanity-docker-milestone
```
GomathiselviS, line 456: Why don't we add milestone branch to the GHA?

mariolenz: I'm not an expert on this GHA stuff, but as far as I understand it's not that easy to make the GHA sanity tests non-voting on GHA. Having this as non-voting won't prevent us from merging PRs but would also give us some information that we might run into problems in the future.

On the other hand, I think we also can remove this test for now even without moving it to GHA. If you want me to, I'll update my PR accordingly. Or you do, because it's time fore me to go to bed...

mariolenz: I've pushed a new commit removing the milestone sanity tests. We can add it later to the GHA workflows in the collection if we want, but let's remove it from Zuul.

mariolenz:
@GomathiselviS What do you think, is this OK for you? We can decide to add milestone and / or devel sanity tests to GHA in the collection later.

jillr (approved):

PR #1846, opened 2024-01-25
Try to override third-party-check config from publish-to-galaxy-3pci

felixfontein:


jillr (approved):

felixfontein:
Hmm, the result is that `noop` was *added* as a new step to `third-party-check`: https://ansible.softwarefactory-project.io/zuul/status/change/784,77b42359f973b8776b360e7ed9d381645eae72ab

jillr:
@felixfontein do you want me to revert?

felixfontein:
@jillr I wanted to try some more to fix this, but I never found time to really dwell into how the Zuul config works. I'll now try something else, and started reverting this (and another PR) in #1859.

PR #1856, opened 2024-04-04
Remove aws nightly job

gravesm:


gravesm:
recheck

tools/zuul-ansible-manager.py
```diff
@@ -649,7 +649,7 @@ def never_called_playbooks() -> bool:
             aws_periodical_jobs(args.amazon_aws_repo_dir)
```
GomathiselviS, line 649: Should we remove this line?

gravesm: Oh, good catch.

GomathiselviS (approved):

PR #1860, opened 2024-05-06
Pin pytest to 8.1.1 in repo CI

felixfontein:


tox.ini
```diff
@@ -21,7 +21,7 @@ commands = {posargs}
 [testenv:pytest]
 deps =
   ansible-core
-  pytest
+  pytest==8.1.0
```
jillr, line 24:
```suggestion
  pytest==8.1.1
```
Looks like 8.1.0 was yanked.

jillr: It's probably fine since zuul is pulling from repo mirrors (that I still haven't been able to find the source of) but I'd prefer to avoid any possible issues with the version that led to it being yanked.

jillr (approved):

felixfontein:
@jillr thanks for reviewing and merging!

PR #1863, opened 2024-05-23
Sanity tests Community.okd

abikouo:
Update `kubernetes.core` version when running job `build-ansible-collection`.
When running sanity tests on branch `X`, the `kubernetes.core` version will be checkout on version `X`
e.g: `community.okd:main` -> `kubernetes.core:main`, `community.okd:Stable-X` -> `kubernetes.core:Stable-X``

### DIFF
--- a/zuul.d/project-templates.yaml
+++ b/zuul.d/project-templates.yaml
@@ -134,7 +134,6 @@
         - build-ansible-collection:
             required-projects:
               - name: github.com/ansible-collections/kubernetes.core
-                override-checkout: stable-2.4
         - ansible-test-sanity-docker-devel
         - ansible-test-sanity-docker-milestone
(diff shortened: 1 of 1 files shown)


gravesm (commented):
We should be testing ansible versions 2.15+. 2.14 is now EOL.

abikouo:
> We should be testing ansible versions 2.15+. 2.14 is now EOL.

@gravesm The intent is to remove `2.14` once the `community.okd:4.0.0` dropping support for `ansible-core<2.15.0` has been published

gravesm:
I'm not sure I understand. Why are we testing new releases against ansible 2.14?

gravesm (commented):
We should be testing on all currently supported versions which is 2.15, 2.16 and 2.17.

PR #1876, opened 2024-06-24
remove netcommon jobs

roverflow:


NilashishC (approved):

NilashishC:
recheck

roverflow:
recheck

mariolenz:
@roverflow I know that the `mergeit` label won't merge anything as long as the CI fails. But

1. someone approves (although the CI is still running!)
2. someone adds the `mergeit` label
3. CI fails
4. someone does a quick'n'dirty change to make the CI succeed
5. this will be merged, nobody reviews this anymore; there's simply not enough time to do it

This looks like an accident waiting to happen. The problem here is that this PR has not only been approved although the CI failed, it's also that any quick'n'dirty workaround to make the CI succeed will be automatically merged _without anyone else having another look at it_.

roverflow:
@mariolenz I understand your concern and agree that this process could lead to potential issues. In this specific case, we had thoroughly discussed the changes and did not anticipate a node failure, which occurred due to Zuul being updated. This is why the mergeit label was added before the CI completed. Usually, we do follow a much stricter process to avoid such situations. We'll ensure to be more cautious in the future.

NilashishC:
recheck

PR #1878, opened 2024-06-25
Remove required-projects for amazon.aws integration jobs

GomathiselviS:
This PR removes the dependencies from required-projects. These will be added to the collection's galaxy.yml.


abikouo:
It seems like this needs to be updated https://github.com/ansible/ansible-zuul-jobs/blob/master/tools/zuul-ansible-manager.py

PR #1894, opened 2025-03-02
Restrict ncclient version in test-requirements.txt

mariolenz:
The last ncclient release [0.6.18](https://pypi.org/project/ncclient/0.6.18/) seems to break the Zuul jobs. ansible-collections/community.vmware#2337 fails with:

```
2025-03-02 12:33:59.304660 | controller | Collecting ncclient
2025-03-02 12:33:59.304669 | controller |   Downloading ncclient-0.6.18.tar.gz (112 kB)
2025-03-02 12:33:59.304678 | controller |      â”â”â”â”â”â”â”â”â”â”â”â”â”â”â”â”â”â”â”â”â”â”â”â”â”â”â”â”â”â”â”â”â”â”â”â”â”â” 112.4/112.4 kB 33.3 MB/s eta 0:00:00
2025-03-02 12:33:59.304687 | controller |   Preparing metadata (setup.py): started
2025-03-02 12:33:59.304696 | controller |   Preparing metadata (setup.py): finished with status 'error'
2025-03-02 12:33:59.304705 | controller |   error: subprocess-exited-with-error
2025-03-02 12:33:59.304714 | controller |
2025-03-02 12:33:59.304726 | controller |   Ã— python setup.py egg_info did not run successfully.
2025-03-02 12:33:59.304735 | controller |   â”‚ exit code: 1
2025-03-02 12:33:59.304744 | controller |   â•°â”€> [7 lines of output]
2025-03-02 12:33:59.304753 | controller |       Traceback (most recent call last):
2025-03-02 12:33:59.304762 | controller |         File "<string>", line 2, in <module>
2025-03-02 12:33:59.304772 | controller |         File "<pip-setuptools-caller>", line 34, in <module>
2025-03-02 12:33:59.304783 | controller |         File "/tmp/pip-install-gc5p4tso/ncclient_037de0269b5b456b80b8306a391519e6/setup.py", line 36, in <module>
2025-03-02 12:33:59.304793 | controller |           test_req_lines = [line.strip() for line in open("requirements-test.txt").readlines()]
2025-03-02 12:33:59.304807 | controller |                                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2025-03-02 12:33:59.304817 | controller |       FileNotFoundError: [Errno 2] No such file or directory: 'requirements-test.txt'
2025-03-02 12:33:59.304826 | controller |       [end of output]
2025-03-02 12:33:59.304835 | controller |
[...]

PR #1896, opened 2025-03-02
Drop ncclient from test requirents

mariolenz:
I think that most collections don't need `ncclient`. So it's special and shouldn't be a general requirement.

see also #1894


mariolenz:
ncclient/ncclient#623

mariolenz:
Should be fixed in 0.6.19: ncclient/ncclient#624

PR #1899, opened 2025-03-17
[amazon.aws/community.aws] add non voting job for ansible-core 2.19

abikouo:


gravesm (commented):
Are we sure we want to do this? If I'm understanding this change correctly, it's going to double the amount of zuul jobs per PR. This is a useful thing to have for the rare one-off feature to test like data tagging, but I'm not convinced this is worth it for day-to-day CI.

PR #1901, opened 2025-03-31
community.vmware: Make ansible-galaxy-importer non-voting

mariolenz:
It looks like `ansible-galaxy-importer` has problems with doc fragments from another collection. So I think we should make it non-voting and test this with a GHA.

At least until it's fixed or I find a way to make it work.

FYI I ran into this when trying to re-use doc fragments from `vmware.vmware` in `community.vmware`: ansible-collections/community.vmware#2369

### DIFF
--- a/zuul.d/project-templates.yaml
+++ b/zuul.d/project-templates.yaml
@@ -137,10 +137,18 @@
         - ansible-test-cloud-integration-vcenter7_2esxi-stable218
         - ansible-test-cloud-integration-vcenter7_1esxi-stable218_1_of_2
         - ansible-test-cloud-integration-vcenter7_1esxi-stable218_2_of_2
+        - ansible-galaxy-importer:
+            voting: false
+    third-party-check:
+      jobs:
+        - ansible-galaxy-importer:
+            voting: false
     gate:
       jobs:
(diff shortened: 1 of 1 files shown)


mariolenz:
> So I think we should make it non-voting and test this with a GHA.

Actually, I think we're already testing this with GHA: https://github.com/ansible-collections/community.vmware/blob/main/.github/workflows/import-galaxy.yml

So `ansible-galaxy-importer` is not only problematic, it's also redundant :smiley:

@alinabuzachis What do you think about disabling this? As far as I understood @mikemorency you've made the job non-voting in `community.aws` for similar reasons.

alinabuzachis (approved):

mariolenz:
> BTW we're already testing this with GHA: https://github.com/ansible-collections/community.vmware/blob/main/.github/workflows/import-galaxy.yml

@mikemorency Do you have an idea who I could ping to get this merged now that it's approved?

mikemorency:
@gravesm helped merge my PR, maybe they could help with this one too

gravesm (approved):

mariolenz:
Thanks @gravesm!

PR #1913, opened 2025-05-23
community.vmware: Drop vmware.vmware_rest requirement

mariolenz:
The modules we use in `community.vmware` to prepare the integration tests are deprecated. So it's a good idea to get rid of them. And then we wouldn't require `vmware.vmware_rest` anymore.

This might also make it possible to undo ansible/zuul-config#587 / redo ansible/zuul-config#586.

### DIFF
--- a/zuul.d/project-templates.yaml
+++ b/zuul.d/project-templates.yaml
@@ -127,10 +127,6 @@
       fail-fast: true
       jobs:
         - ansible-tox-linters
-        - build-ansible-collection:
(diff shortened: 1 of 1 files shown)


mariolenz:
Thanks a lot @jillr!

mariolenz:
Thanks again @jillr! For the record:

* ansible-collections/community.vmware#2388
* ansible-collections/community.vmware#2400

PR #1916, opened 2025-08-16
community.vmware: Drop vSphere 7 integration tests

mariolenz:
Since vSphere 7 will be EOL soon, I think we should remove the integration tests. Especially since 1) the vCenter license has expired again and 2) the people who have maintained this so far struggle to continue doing so (ansible-collections/community.vmware#1746). Personally, I don't know how to get a new vCenter image up and running in Zuul, and if I did I think I wouldn't have the permissions to do it.

So I think it would be best to remove those integration tests. Otherwise, all PRs in `comunity.vmware` will fail because of the expired test license. I still don't know how to provide an alternative. But not being able to merge PRs is even worse than not having integration tests with a real vSphere environment.

cc @ihumster


mariolenz:
Thanks @jillr!

gundalow:
> Especially since 1) the vCenter license has expired again and 2) the people who have maintained this so far struggle to continue doing so (https://github.com/ansible-collections/community.vmware/issues/1746).

I accept responsibility for not ensuring the change in tested was communicated at all, or in advance of the change.

@mariolenz Thank you for updating the testing.

